PR #61, opened 2022-07-31
Add Missing SSH Certificate Types

stevenelleman-okta:
Add missing types and refactor / clean up the pattern.


docs/resources/project.md
```diff
@@ -28,7 +28,7 @@ An ASA construct that contains servers and is used to grant end users access to
...
+- `ssh_certificate_type` (String) The SSH certificate type used by access requests. Options include: [`CERT_TYPE_ECDSA_521_01`, `CERT_TYPE_ECDSA_384_01`, `CERT_TYPE_ECDSA_256_01`, `CERT_TYPE_ED25519_01`, `CERT_TYPE_RSA_SHA2_512_01`, `CERT_TYPE_RSA_SHA2_256_01`, `CERT_TYPE_RSA_01`]. `CERT_TYPE_RSA_01` is a deprecated key algorithm type. This option should only be used to connect to legacy systems that cannot use newer SSH versions. If you do need to use `CERT_TYPE_RSA_01`, it is recommended to connect via a gateway with traffic forwarding. Otherwise, please use a more current key algorithm. If left unspecified, `CERT_TYPE_ED25519_01` is used by default.
```
waltergoulet-okta, line 31: Nit, but since CERT_TYPE_ED25519_01 is our default it feels like it would be better if this was the first item in this list.

oktapam/constants/typed_strings/key_algos.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"fmt"
+	"strings"
+)
+
+type SSHCertificateType string
+
+const (
+	CertTypeRsa      = "CERT_TYPE_RSA_01"
+	CertTypeRsa256   = "CERT_TYPE_RSA_SHA2_256_01"
```
waltergoulet-okta, line 12: The API docs don't list variants of RSA sig algorithm supported beyond plain CERT_TYPE_RSA_01. See https://developer.okta.com/docs/reference/api/asa/projects/#request-body-2. Are our API docs inaccurate and don't reflect what we support or are these algorithms added here not supported by our backend?

stevenelleman-okta: So it looks like they are options we [support](https://github.com/okta/platform/blob/master/backend/models/project.go#L57) but not in the [UI](https://github.com/okta/platform/pull/5792) -- should they be added to the UI?

waltergoulet-okta: It goes beyond UI though because our API docs also don't show those other RSA sig options. I'm inclined to leave it as is because a) if we show them through UI we need to update API docs and b) we are in general trying to deprecate use of RSA (even though those other options are more secure versions of RSA sig alg). It should be a separate feature request (unrelated to TF provider) to 'productize' those other options.

waltergoulet-okta (approved):

PR #65, opened 2022-08-02
Added tests for AD Connection, Task and Data source

sachinsaxena-okta:
**CHANGELOG:** Added acceptance tests for AD connection resources & DS

**Code Changes**

1. Added acceptance tests for AD Connection Resource, AD Task Settings Resource & AD Connections Datasource
2. Renamed AD Connections DS filename
3. Made "Run Test" attribute in AD Task Settings both optional & computed.


examples/oktapam_ad_connection/ad_connection_example.tf
```diff
@@ -13,14 +13,14 @@ resource "oktapam_project" "test-project" {
 }
 
 # Creating gateway is not supported. Get the gateway id using datasource
-data "oktapam_gateway" "gateway-list" {
```
stevenelleman-okta, line 16: James recommended all names be kept consistent with underscores, e.g. `gateway_list`

sachinsaxena-okta: Fixed

stevenelleman-okta:
Need to fix test.

sachinsaxena-okta:
> Need to fix test.

I fixed the test and waiting for populating project id decision. To fix tests, project id is required.

sachinsaxena-okta:
Added project id attribute to project resource.

oktapam/client/project.go
```diff
@@ -37,7 +37,7 @@ func (p Project) ToResourceMap() map[string]interface{} {
 		m[attributes.Name] = *p.Name
 	}
 	if p.ID != nil {
-		m[attributes.ID] = *p.ID
+		m[attributes.ProjectID] = *p.ID
```
jamesatwill-okta, line 40: I don't think I understand this. Shouldn't this map to the resource's ID?

oktapam/resource_ad_task_settings.go
```diff
@@ -30,16 +34,11 @@ func resourceADTaskSettings() *schema.Resource {
 				Required:    true,
 				Description: descriptions.Name,
 			},
-			attributes.ID: {
```
jamesatwill-okta, line 33: I think I might just not understand IDs?

oktapam/client/project.go
sachinsaxena-okta: Terraform Resource Id in project resource is set to "Project Name". Earlier, when I discussed with Steven the idea was to set it to some unique identifier which is used to retrieve the resource back from the data store via apis. So, ASA project id was not set as any attribute. But, we need "ASA Project Id" while creating resources like "AD Task Settings". We can pass name too but then it will require one extra "Get Project by name" call to get the project id.

**Concatenating multiple resource Attribute Pattern** -  In few cases, we are concatenating ASA Identifiers even though we already have required attributes in the resource schema. For Ex - For Project Group resource, pattern for Terraform resource id is <Project Name>:<Group Name>. Project Group resource schema also has attributes for Project and Group name. 

To keep Id attribute consistent, my proposal is to set it to ASA resource Id. We can always retrieve the resource back via API as long as required attributes are set in resource schema. For that purpose, I created [PR](https://github.com/okta/terraform-provider-oktapam/pull/69). I didn't change it currently for all the resources as I first want to collect feedback and see if we agree with the proposal.

oktapam/resource_ad_task_settings.go
sachinsaxena-okta: ID is very confusing in the terraform. Id is treated as "meta-field" and it has it's own getter/setter. 

(*helper/schema.ResourceData).SetId() method does two things - 

- Set the underlying resource identifier
- Set the resource id attribute. 

Existence of resource identifier determine if resource exists or not and it's mandatory to set it while creating resource. I was not sure what is the purpose of keeping ID as the attribute. Then, I looked at the terraform sdk code and see below comment - 

	// once we transition away from the legacy state types, "id" will no longer
	// be a special field, and will become a normal attribute.

Reference: https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/schema/resource_data.go#L286

I can make a change and keep the Id attribute.

jamesatwill-okta: Ah so if I understand correctly, us humans should not really be using "id" - at least not for project, since if we want the project's id we should use `project_id`, so having an `id` field too is confusing because it's actually the project's name?

sachinsaxena-okta: Sorry, my comment wasn't very clear. I was just trying to explain how "Id" attribute work in terraform and how it's going to change over time and not treated as "meta-field".This PR was created before putting thoughts around how we want to populate tf resource id, so at that time I created extra attribute "project_id" to make tests work. Now, I feel there is no need to introduce new attribute and we can just populate existing id attribute. 

My proposal is to populate "Id" as ASA resource UUID wherever it exist (for plural datasource there is no ASA resource Id). ASA resource UUID is unique and stable attribute. We don't need to base our decision to populate "id" on how read resource api works. We can change tf resource read method and import method to use other required attributes instead of id to retrieve resource back. @stevenelleman-okta want to review this approach first with you before making changes all over. 

I think better to talk on this once you come online.

oktapam/resource_ad_task_settings.go
```diff
@@ -30,16 +34,11 @@ func resourceADTaskSettings() *schema.Resource {
...
-				Type:     schema.TypeString,
-				Computed: true,
```
jamesatwill-okta, line 35: So from our chat, this ID comes back.

oktapam/resource_ad_task_settings.go
```diff
@@ -181,10 +181,12 @@ func resourceADTaskSettingsRead(ctx context.Context, d *schema.ResourceData, m i
 	var diags diag.Diagnostics
 	c := m.(client.OktaPAMClient)
 
-	id := d.Id()
-	adConnId := d.Get(attributes.ADConnectionID).(string)
+	adConnId, adTaskSettingsId, err := parseADTaskSettingsResourceID(d.Id())
```
jamesatwill-okta, line 184: So this can have a d[conneciton_id], and d.Id() is just ad task settings ID?

oktapam/resource_ad_task_settings.go
```diff
@@ -194,16 +196,19 @@ func resourceADTaskSettingsRead(ctx context.Context, d *schema.ResourceData, m i
...
 
 	return diags
 }
 
 func resourceADTaskSettingsUpdate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
 	c := m.(client.OktaPAMClient)
-	taskSettingsId := d.Id()
-	adConnId := d.Get(attributes.ADConnectionID).(string)
+
+	adConnId, adTaskSettingsId, err := parseADTaskSettingsResourceID(d.Id())
```
jamesatwill-okta, line 208: Same here (and s/Id/ID/)

oktapam/resource_ad_task_settings.go
sachinsaxena-okta: Fixed in [6c0355a](https://github.com/okta/terraform-provider-oktapam/pull/65/commits/6c0355a9175e980f2658c6d23769aa347eac8d8f)

oktapam/resource_ad_task_settings.go
sachinsaxena-okta: Added Id back. Fixed in [6c0355a](https://github.com/okta/terraform-provider-oktapam/pull/65/commits/6c0355a9175e980f2658c6d23769aa347eac8d8f)

oktapam/resource_ad_task_settings.go
sachinsaxena-okta: Done. Fixed in [6c0355a](https://github.com/okta/terraform-provider-oktapam/pull/65/commits/6c0355a9175e980f2658c6d23769aa347eac8d8f)

oktapam/resource_ad_task_settings.go
sachinsaxena-okta: Fixed in [6c0355a](https://github.com/okta/terraform-provider-oktapam/pull/65/commits/6c0355a9175e980f2658c6d23769aa347eac8d8f)

oktapam/client/project.go
sachinsaxena-okta: Removed project id attribute and populated ASA UUID for "id" attribute instead of name.

jamesatwill-okta (approved):

PR #72, opened 2022-08-23
Use Github Action to Set Up Terraform CLI

stevenelleman-okta:
Every acceptance test run the terraform-cli is downloaded from https://checkpoint-api.hashicorp.com/v1/check/terraform?arch=amd64&os=linux&signature=&version=

Periodically acceptance tests fail when the download times out.

This PR uses the terraform setup github action in the acceptance tests, let's see if it addresses this class of failure.


stevenelleman-okta:
I don't love the terraform cli versions peppered around, I'm looking into variable usage in Github Actions.

andrewhamilton-okta (approved):

PR #73, opened 2022-08-29
Updates terraform binary provisioning

andrewhamilton-okta:
- Updates the terraform bin setup script to pull down the correct
  version.
- Provides the local terraform binary to the running Docker container
  with the integration tests.
- Turns off usage of the terraform wrapper since we don't need it and
  adds extra dependencies.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -22,13 +22,6 @@ jobs:
     steps:
       - name: Check out repository code
         uses: actions/checkout@v3
-      - name: Get Terraform CLI Version
-        id: vars
-        run: echo ::set-output name=tf-version::$(cat .terraform-version)
-      - name: Setup Terraform
-        uses: hashicorp/setup-terraform@v2
-        with:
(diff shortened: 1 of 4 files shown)


.github/workflows/checks.yml
```diff
@@ -22,13 +22,6 @@ jobs:
     steps:
       - name: Check out repository code
         uses: actions/checkout@v3
-      - name: Get Terraform CLI Version
-        id: vars
-        run: echo ::set-output name=tf-version::$(cat .terraform-version)
-      - name: Setup Terraform
-        uses: hashicorp/setup-terraform@v2
-        with:
-          terraform_version: ${{ steps.vars.outputs.tf-version }}
```
stevenelleman-okta, line 31: I think there may too more of these to remove, just search "Get Terraform CLI Version".

andrewhamilton-okta: @stevenelleman-okta We just don't need it for the unit tests since those are just directly on the provider Go code. We do need it for the acceptance tests still which are the other two entries.

stevenelleman-okta (approved):

PR #78, opened 2022-09-15
Fix create_server_group Attribute on Project Group

stevenelleman-okta:


wayneearl-okta (approved):
Tested and confirmed that this change corrects the issue.

PR #79, opened 2022-09-15
Fix ProjectGroup server_selectors Behavior

stevenelleman-okta:
When `server_selectors` value is updated in the dashboard UI there are three cases to consider, and one of the three cases is failing. 

This works correctly for:
- Config: Value “A”, Actual: Value “B”, Apply action: “B” > “A”
- Config: None, Actual: Value “B”, Apply action: “B” > null

However, this scenario does not work correctly:
- Config: Value “A”, Actual: None, Apply action: N/A

`ProjectGroup.ToResourceMap()` skipped `server_selectors` if it was empty. I suspect that if the value is omitted, Terraform retains the current value of that attribute. Now it'll be set if empty, and the Terraform state should be updated in that third case.


wayneearl-okta (approved):
Tested and confirmed that this change fixes the issues.

PR #82, opened 2022-09-29
Fix ssh_certificate_type not set on project update

WojoInc:
## Issue

Currently when updating a project, changing the SSH Certificate Type used by the project results in a plan that
appears to change the certificate type, but a subsequent apply does change this attribute on the project. 
This is verifiable in ASA console, and by running a subsequent call to `terraform plan`. 

Seeing as this attribute is changeable in the console, I believe it makes sense to be able to do so from the provider as well.

## Change Summary 

Adds SSHCertificateType to the list of attributes that are modifyable during a project update.

Specifically:

- Adds missing `attributes.SSHCertificateType` to the slice of `changeableAttributes` in `resourceProjectUpdate` function
- Updates test cases to account for a change in certificate type

## Notes

Verified that all acceptance tests pass when run against my dev tenant (with the exception of user tests, which have the tenant/team hardcoded)

I chose `CERT_TYPE_RSA_01` when updating the test cases, but I'm sure that could be changed to a different cert type if preferred.

### DIFF
--- a/oktapam/resource_project.go
+++ b/oktapam/resource_project.go
@@ -237,6 +237,7 @@ func resourceProjectUpdate(ctx context.Context, d *schema.ResourceData, m interf
 		attributes.SSHSessionRecording,
 		attributes.GatewaySelector,
 		attributes.UserOnDemandPeriod,
+		attributes.SSHCertificateType,
 	}
 
 	for _, attribute := range changeableAttributes {
--- a/oktapam/resource_project_test.go
+++ b/oktapam/resource_project_test.go
@@ -39,7 +39,7 @@ func TestAccProject(t *testing.T) {
 		RDPSessionRecording:    utils.AsBoolPtrZero(true, true),
 		SSHSessionRecording:    utils.AsBoolPtrZero(true, true),
 		GatewaySelector:        utils.AsStringPtr("env=test"),
-		SSHCertificateType:     utils.AsStringPtr("CERT_TYPE_ED25519_01"),
+		SSHCertificateType:     utils.AsStringPtr("CERT_TYPE_RSA_01"),
 		UserOnDemandPeriod:     utils.AsIntPtr(10),
 	}
 	resource.Test(t, resource.TestCase{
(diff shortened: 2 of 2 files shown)


oktapam/resource_project.go
```diff
@@ -237,6 +237,7 @@ func resourceProjectUpdate(ctx context.Context, d *schema.ResourceData, m interf
 		attributes.SSHSessionRecording,
 		attributes.GatewaySelector,
 		attributes.UserOnDemandPeriod,
+		attributes.SSHCertificateType,
 	}
```
stevenelleman-okta, line 241: Good catch.

oktapam/resource_project_test.go
```diff
@@ -39,7 +39,7 @@ func TestAccProject(t *testing.T) {
 		RDPSessionRecording:    utils.AsBoolPtrZero(true, true),
 		SSHSessionRecording:    utils.AsBoolPtrZero(true, true),
 		GatewaySelector:        utils.AsStringPtr("env=test"),
-		SSHCertificateType:     utils.AsStringPtr("CERT_TYPE_ED25519_01"),
```
stevenelleman-okta, line 42: Any reason the type was changed?

WojoInc: I think if I'm reading the test case right, L42 needs to match L85 so that the test case passes. I picked something other than `CERT_TYPE_ED25519_01` so that the test would actually check that the value was updated. I can definitely pick something other than RSA_01 though.

Changing it back results in a failure when running the acceptance tests:
```log
TESTARGS="-run TestAccProject" make testacc
TF_ACC=1 go test ./... -v -run TestAccProject -timeout 120m   
?       github.com/okta/terraform-provider-oktapam      [no test files]
=== RUN   TestAccProjectGroup
--- PASS: TestAccProjectGroup (13.34s)
=== RUN   TestAccProject
    resource_project_test.go:45: Step 2/3 error: Check failed: 1 error occurred:
                * Check 1/6 error: expected project does not match returned project.
         {
          Name: "test_acc_project_MGTNLLDX1Q",
          ID: "1ab087a9-7f9c-4dcf-901d-446a3e048792",
          Team: "wojoinc",
          DeletedAt: nil,
          NextUnixGID: 63400,
          NextUnixUID: 61200,
          CreateServerUsers: true,
          ForwardTraffic: true,
          RDPSessionRecording: true,
          RequirePreAuthForCreds: false,
          SSHSessionRecording: true,
          GatewaySelector: "env=test",
        - SSHCertificateType: "CERT_TYPE_RSA_01",
        + SSHCertificateType: "CERT_TYPE_ED25519_01",
          UserOnDemandPeriod: 10,
         }
        
--- FAIL: TestAccProject (9.14s)
```

stevenelleman-okta: That reason is fine, just curious.

WojoInc: Okay sounds good!

stevenelleman-okta (approved):

PR #83, opened 2022-10-12
OKTA-541126 - set the default create_server_users value to be true

SajanAlexander-okta:
The ASA UI currently sets the `create_server_users` value to be true as a default, so we should be in line with that.  The setting of false does not allow for us to manage the life cycle of accounts on the server, so it is an important one to respect the default of true.


sachinsaxena-okta (approved):
Looks good

PR #85, opened 2022-11-03
Fixed Ad connections test

sachinsaxena-okta:
AD Connections test were failing because ad connection name and domain name prefixes used in tests were not unique. One more issue was querying data source by GatewayId. Gateway is not a TF managed resource and test to check no of connections by gateway id will give inconsistent results. Changed the test for now to query by certificate id.

### DIFF
--- a/oktapam/data_source_ad_connections_test.go
+++ b/oktapam/data_source_ad_connections_test.go
@@ -9,26 +9,24 @@ import (
 )
 
 func TestAccDatasourceADConnections(t *testing.T) {
-	adConnectionNamePrefix := fmt.Sprintf("test_acc_datasource_ad_connection_%s", randSeq(10))
-	adConnectionName1 := fmt.Sprintf("%s_1", adConnectionNamePrefix)
-
-	adConnectionResNamePrefix := "oktapam_ad_connection.test_acc_ad_connection"
-	adConnectionResName1 := fmt.Sprintf("%s_1", adConnectionResNamePrefix)
-
-	resourceNamePrefix := "data.oktapam_ad_connections.test_acc_datasource_ad_connections"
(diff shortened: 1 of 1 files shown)


oktapam/data_source_ad_connections_test.go
```diff
@@ -39,21 +37,31 @@ const testAccDatasourceADConnectionsConfigFormat = `
...
-	 use_passwordless         = false
+	 use_passwordless         = true
+     certificate_id  		  = oktapam_ad_certificate_request.self_signed_cert.id
 }
 
-data "oktapam_ad_connections" "test_acc_datasource_ad_connections_1" {
-    gateway_id = data.oktapam_gateways.gateways.gateways[0].id
-    depends_on = [oktapam_ad_connection.test_acc_ad_connection_1]
+data "oktapam_ad_connections" "test_acc_datasource_ad_connections" {
+    certificate_id = oktapam_ad_certificate_request.self_signed_cert.id
```
stevenelleman-okta, line 60: Did this change too?

oktapam/data_source_ad_connections_test.go
```diff
@@ -39,21 +37,31 @@ const testAccDatasourceADConnectionsConfigFormat = `
...
+  type         = "self_signed"
+  details {
+	ttl_days = 1
+  }
+}
+
+resource "oktapam_ad_connection" "test_acc_ad_connection" {
 	 name                     = "%[1]s"
 	 gateway_id               = data.oktapam_gateways.gateways.gateways[0].id
-	 domain                   = "testacc.example.com"
```
stevenelleman-okta, line 45: Ah gotcha, so the fact this was the same domain every time meant that the tests could clash?

stevenelleman-okta (approved):

PR #87, opened 2022-11-11
Increase Entropy of Test Names to Avoid Collisions

stevenelleman-okta:


oktapam/provider_test.go
```diff
@@ -47,9 +47,9 @@ func testAccPreCheck(t *testing.T) {
 
 var randChars = []rune("ABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890")
 
-func randSeq(n int) string {
+func randSeq() string {
 	r := rand.New(rand.NewSource(time.Now().UnixNano()))
-	b := make([]rune, n)
```
stevenelleman-okta, line 52: Opted to change the character length globally. In certain contexts this may be overkill but I figured too much entropy is better than too little entropy (unless some kind of name character limit is exceeded, but we'll figure it out on the first test run).

oktapam/provider_test.go
```diff
@@ -47,9 +47,9 @@ func testAccPreCheck(t *testing.T) {
...
+	b := make([]rune, 20) // Character length
```
stevenelleman-okta, line 52: Opted to change the character length globally. In certain contexts this may be overkill but I figured too much entropy is better than too little entropy (unless some kind of name character limit is exceeded, but we'll figure that out on the first test run).

SajanAlexander-okta (approved):

PR #88, opened 2022-11-29
Add Retry to Failed Acceptance Tests 

stevenelleman-okta:
There are at least two common classes of non-deterministic acceptance test failures: dependency download timeouts and failed locks (from tests being run in parallel). 

The intention of a retry is to reduce false positives.


jamesatwill-okta (approved):
Gross, but until we get time to actually cache these somewhere reliable, hopefully it reduces the false negatives we've been getting.

PR #95, opened 2023-03-02
OKTA-556100 terraform changes to use user sync task settings

asitchakraborti-okta:
Adds user sync task settings support to oktapam terraform provider, includes tests and data sources.


oktapam/constants/descriptions/links.go
```diff
@@ -5,6 +5,7 @@ const (
 	LinkADConnection          = "For more information check out the [documentation](https://help.okta.com/asa/en-us/Content/Topics/Adv_Server_Access/docs/ad-connections.htm) on AD Connection creation and usage."
 	LinkADRuleAssignment      = "For more information check out the [documentation](https://help.okta.com/asa/en-us/Content/Topics/Adv_Server_Access/docs/ad-sync-assignment.htm) on AD rule assignment."
 	LinkADServerDiscovery     = "For more information check out the [documentation](https://help.okta.com/asa/en-us/Content/Topics/Adv_Server_Access/docs/ad-sync.htm) on AD server discovery."
+	LinkADUserDiscovery       = "For more information check out the [documentation](https://help.okta.com/asa/en-us/Content/Topics/Adv_Server_Access/docs/ad-sync.htm) on AD user discovery."
 	LinkGateway               = "For details, see [Advanced Access Server Gateways](https://help.okta.com/asa/en-us/Content/Topics/Adv_Server_Access/docs/gateways-and-bastions.htm) on ASA Gateways and bastions."
```
asitchakraborti-okta, line 9: Will need followup.

oktapam/resource_ad_task_settings.go
```diff
@@ -357,7 +357,7 @@ func flattenADTaskSettings(taskSettings *client.ADTaskSettings) map[string]inter
 		m[attributes.IsActive] = *taskSettings.IsActive
 	}
 	if taskSettings.RunTest != nil {
-		m[attributes.RunTest] = *taskSettings.IsActive
+		m[attributes.RunTest] = *taskSettings.RunTest
```
asitchakraborti-okta, line 360: Fixing this bug, separate form this, RunTest might need more consideration as a standalone RunTest change does not make sense and is skipped anyway, but should not produce any noise with plan etc.

oktapam/constants/descriptions/links.go
sachinsaxena-okta: Talk to Walter on this, we can have something similar to Kubernetes Cluster and mark this as Beta Feature.

oktapam/client/ad_connection.go
```diff
@@ -346,3 +368,161 @@ func (p ListADConnectionsParameters) toQueryParametersMap() map[string]string {
...
+	if _, err := checkStatusCode(resp, http.StatusCreated); err != nil {
+		logging.Tracef("unexpected status code: %d", resp.StatusCode())
+		return nil, err
+	}
+
+	createdADUserSyncTaskSettings := resp.Result().(*ADUserSyncTaskSettings)
+	return createdADUserSyncTaskSettings, nil
+}
+
+// TODO can we get rid of this
```
sachinsaxena-okta, line 452: Yes we can get rid of this.

oktapam/resource_ad_user_sync_task_settings.go
```diff
@@ -0,0 +1,231 @@
...
+				ForceNew:    true,
+				Description: descriptions.ADUserSyncTaskSettingsBaseDN,
+			},
+			attributes.LDAPQueryFilter: {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: descriptions.ADUserSyncTaskSettingsLDAPQueryFilter,
+			},
+			attributes.UPNField: {
```
sachinsaxena-okta, line 72: This field should be marked as optional. I think at database end we can set default. Check how it's handled for AdTaskSettings -> HostNameAttribute

oktapam/resource_ad_user_sync_task_settings.go
```diff
@@ -0,0 +1,231 @@
...
+				Required:    true,
+				ForceNew:    true,
+				Description: descriptions.ADUserSyncTaskSettingsSIDField,
+			},
+			attributes.IsActive: {
+				Type:        schema.TypeBool,
+				Required:    true,
+				Description: descriptions.ADUserSyncTaskSettingsIsActive,
+			},
+			attributes.RunTest: {
```
sachinsaxena-okta, line 89: I am going to check more on this. I think the ideal way to handle it is to suppress diff for this attribute. This attribute is only required while creating resource and then it's never returned by the api while reading it back. Computed is used for attributes which are not known at creation time and returned back after creating the resource for example resource identifiers.

oktapam/resource_ad_user_sync_task_settings.go
asitchakraborti-okta: Task settings use those as Required/Forcenew, but may be because there the API does not have default. I can try optional here, we have to deal with the update semantics, I guess we can say missing attribute there means use default (and not previous values).

oktapam/resource_ad_user_sync_task_settings.go
```diff
@@ -0,0 +1,231 @@
...
+				ForceNew:    true,
+				Description: descriptions.ADUserSyncTaskSettingsUPNField,
+			},
+			attributes.SIDField: {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: descriptions.ADUserSyncTaskSettingsSIDField,
+			},
+			attributes.IsActive: {
```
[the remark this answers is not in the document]
asitchakraborti-okta: Here the update semantics will be problematic

oktapam/constants/descriptions/links.go
sachinsaxena-okta: We can update this link later once we have documentation available.

asitchakraborti-okta: opened https://oktainc.atlassian.net/browse/OKTA-593282

oktapam/resource_ad_user_sync_task_settings.go
asitchakraborti-okta: ok

oktapam/resource_ad_user_sync_task_settings.go
asitchakraborti-okta: Can't find a primitive that does "ignore this attribute for update trigger calculation, but use it as it is specified for create". There is DiffSuppressFunc, which can avoid triggering update, but will use the false value always when we recreate. Note that at present the only problem is the noise during plan when the value is true, ideally it should not be there, and the execution should probably not enter provider logic, but at least no platform API call is made in that case.

oktapam/resource_ad_user_sync_task_settings.go
```diff
@@ -0,0 +1,230 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: descriptions.Name,
+			},
+			attributes.ID: {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			attributes.Frequency: {
```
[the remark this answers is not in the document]
asitchakraborti-okta: Tried the original task_settings APIs. There is no restriction on frequency, a frequency of 7 will result in */7, looks like at the day juncture the gap will be higher that 7 hours...

oktapam/resource_ad_user_sync_task_settings.go
```diff
@@ -0,0 +1,230 @@
...
+				Type:         schema.TypeInt,
+				Required:     true,
+				ValidateFunc: validation.IntBetween(1, 24),
+				Description:  descriptions.ADUserSyncTaskSettingsFrequency,
+			},
+			attributes.StartHourUTC: {
+				Type:         schema.TypeInt,
+				Optional:     true,
```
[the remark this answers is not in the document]
asitchakraborti-okta: This was following the task_settings pattern, where it is optional. But I guess a validate failure is better in this clearly defined case, will look into the validation.

oktapam/data_source_ad_user_sync_task_settings_list.go
```diff
@@ -0,0 +1,135 @@
...
+				Description: descriptions.FilterConnectionID,
+			},
+			// Return value
+			attributes.ADUserSyncTaskSettingsList: {
+				Type:        schema.TypeList,
+				Computed:    true,
+				Description: descriptions.SourceADUserSyncTaskSettingsList,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						attributes.ADConnectionID: {
```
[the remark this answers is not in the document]
asitchakraborti-okta: will remove

oktapam/resource_ad_user_sync_task_settings.go
```diff
@@ -0,0 +1,230 @@
...
+	} else {
+		//Set returned id
+		d.SetId(*createdADUSTS.ID)
+	}
+
+	return resourceADUserSyncTaskSettingsRead(ctx, d, m)
+}
+
+func resourceADUserSyncTaskSettingsRead(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+	var diags diag.Diagnostics
```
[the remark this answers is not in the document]
asitchakraborti-okta: ok

oktapam/resource_ad_user_sync_task_settings.go
```diff
@@ -0,0 +1,230 @@
...
+
+	if adUserSyncTaskSettings != nil && utils.IsNonEmpty(adUserSyncTaskSettings.ID) {
+		for key, value := range adUserSyncTaskSettings.ToResourceMap() {
+			_ = d.Set(key, value)
+		}
+	} else {
+		logging.Infof("ADUserSyncTaskSettings %s does not exist", adUserSyncTaskSettingsID)
+	}
+
+	return diags
```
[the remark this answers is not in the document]
asitchakraborti-okta: ok

oktapam/resource_ad_user_sync_task_settings.go
```diff
@@ -0,0 +1,230 @@
...
+	adUserSyncTaskSettingsReq := client.ADUserSyncTaskSettings{
+		Name:            getStringPtr(attributes.Name, d, false),
+		Frequency:       getIntPtr(attributes.Frequency, d, false),
+		StartHourUTC:    getIntPtr(attributes.StartHourUTC, d, false),
+		BaseDN:          getStringPtr(attributes.BaseDN, d, false),
+		LDAPQueryFilter: getStringPtr(attributes.LDAPQueryFilter, d, false),
+		UPNField:        getStringPtr(attributes.UPNField, d, false),
+		SIDField:        getStringPtr(attributes.SIDField, d, false),
+		IsActive:        getBoolPtr(attributes.IsActive, d, true),
+		RunTest:         getBoolPtr(attributes.RunTest, d, true),
```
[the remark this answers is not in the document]
asitchakraborti-okta: This is not optional in the API.

oktapam/data_source_ad_user_sync_task_settings_list.go
```diff
@@ -0,0 +1,135 @@
...
+						typed_strings.TaskSettingsStatusInActive.String(),
+					}, false),
+			},
+			attributes.ADConnectionID: {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: descriptions.FilterConnectionID,
+			},
+			// Return value
+			attributes.ADUserSyncTaskSettingsList: {
```
sachinsaxena-okta, line 39: I missed this part in my initial review. For plural data sources, we are just returning IDs only instead of the whole resource object. This standard is followed in most of the tf providers. Check datasource data_source_projects.go for implementation. I know not all of the datasources(gateways, ad_connections) in the oktapam provider following this standard because those were created before. I will create user stories to fix those too so that we have consistent code.

oktapam/resource_ad_user_sync_task_settings_test.go
```diff
@@ -0,0 +1,217 @@
...
+	domainName := fmt.Sprintf("%s.example.com", nameIdentifier)
+
+	//Build required pre-req config. AD User Sync Tasks Settings require AD Connection and Project
+	preConfig := createTestAccADUserSyncTaskSettingsPreConfig(adConnectionName, projectName, domainName)
+
+	//Update schedule
+	updatedFreq := 24
+	updatedStartHourUTC := 6
+
+	resource.Test(t, resource.TestCase{
```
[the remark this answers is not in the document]
asitchakraborti-okta: ok

oktapam/data_source_ad_user_sync_task_settings_list.go
asitchakraborti-okta: ok will follow that pattern

oktapam/resource_ad_user_sync_task_settings.go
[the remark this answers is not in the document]
asitchakraborti-okta: Added customize diff, also made freq optional as well. In both types of task-settings, that is permissible for inactive settings, there doesn't seem to be any reason to block it in TF, particularly for user-sync, which allows more flexible activate/deactivate.

oktapam/resource_ad_user_sync_task_settings.go
```diff
@@ -227,3 +228,21 @@ func importADUserSyncTaskSettingsState(ctx context.Context, d *schema.ResourceDa
 	d.SetId(adUserSyncTaskSettingsID)
 	return []*schema.ResourceData{d}, nil
 }
+
+func resourceADUserSyncTaskSettingsCustomizeDiff(ctx context.Context, diff *schema.ResourceDiff, v interface{}) error {
+	if diff.HasChanges(attributes.Frequency, attributes.StartHourUTC, attributes.IsActive) {
+		Frequency := diff.Get(attributes.Frequency).(int)
+		StartHourUTC := diff.Get(attributes.StartHourUTC).(int)
+		IsActive := diff.Get(attributes.IsActive).(bool)
+		if IsActive && Frequency < 1 {
```
sachinsaxena-okta, line 237: Do we need to say production here or just say "frequency must be specified when scheduling AD User Sync Task"

asitchakraborti-okta: Yes, I will replace production with active

oktapam/constants/descriptions/attributes.go
```diff
@@ -23,6 +23,14 @@ var (
 	ADTaskIsActive                   = "If `true`, enables AD task"
 	ADTaskRunTest                    = "If `true`, test is performed based on specified AD Task Settings"
 	ADTaskStartHourUTC               = "If AD task is scheduled to run daily, then specify start hour in UTC"
+	ADUserSyncTaskSettingsBaseDN          = "Specifies where the rule searches for users."
+	ADUserSyncTaskSettingsFrequency       = "Frequency of the AD User Sync Task"
+	ADUserSyncTaskSettingsIsActive        = "If `true`, enables AD user sync task"
+	ADUserSyncTaskSettingsLDAPQueryFilter = "Specifies the specific criteria used to filter users."
+	ADUserSyncTaskSettingsRunTest         = "If `true`, test is performed based on specified AD User Sync Task Settings"
+	ADUserSyncTaskSettingsSIDField        = "Name of the AD SID field in User Sync queries"
```
sachinsaxena-okta, line 31: May be change this to => AD Attribute mapped to user security identifier that can be used by ASA to include in the passwordless certificate.

sachinsaxena-okta: We can revisit these later once official documentation is posted.

asitchakraborti-okta: Yes, these probably should be in sync with docs.

sachinsaxena-okta (commented):
LGTM. Just have one minor question on test.

sachinsaxena-okta:
You need to generate doc again. Do, make generate so that Doc Generation doesn't fail. Everything else looks good to me.

sachinsaxena-okta (approved):

PR #101, opened 2023-04-14
Removed setting run_test attribute

sachinsaxena-okta:
Don't set Run_Test attribute for AD Task Settings resource while reading the resource back. API always return false for attribute run_test, regardless of what was passed while creating/updating tasks settings resource. If we set this attribute while reading back, then tf going to show drift during plan while comparing it with the previous state (if run_test was set to 'true" initially).

To avoid this, put the solution:
1. While reading resource back, ignore setting run_test. Whatever value is coming as part of tf config (proposed state) will be set in the state and passed to api too while creating/updating resource.
2. If only attribute updated is run_test, don't do anything and just update the provided value in state.

### DIFF
--- a/oktapam/client/ad_connection.go
+++ b/oktapam/client/ad_connection.go
@@ -381,6 +381,11 @@ func (p ListADConnectionsParameters) toQueryParametersMap() map[string]string {
 	return m
 }
 
+// ToResourceMap Convert API Object to flat map for saving in terraform state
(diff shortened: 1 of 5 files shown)


asitchakraborti-okta (approved):
Looks good, the comment in the PR probably could go to code to help with similar situations in future...

sachinsaxena-okta:
That's right. I will add these in the comments.

PR #102, opened 2023-04-17
Skip Tag Test While No Tags to Test

stevenelleman-okta:
0.3.2 uses go v1.17, 0.3.3 changed to using go v1.18. To retest 0.3.2 we would need to run the test using go v1.17. For the time being skip the tag / test. 

Also addresses vuln that was [flagged by dependabot](https://github.com/okta/terraform-provider-oktapam/security/dependabot/13), by updating `golang.org/x/net`.


sachinsaxena-okta (commented):
LGTM

sachinsaxena-okta (approved):

PR #105, opened 2023-06-26
OKTA-615109 - update provider for PAM EA featureset

SajanAlexander-okta:
Large PR for feature-set included in the PAM Early Access.


docs/resources/resource_group.md
```diff
@@ -0,0 +1,28 @@
...
+
+
+
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `delegated_resource_admin_groups` (List of String) The ids of the groups for users who are designated as delegated admins for the resource group
```
waltergoulet-okta: We don't call these objects delegated resource admin groups in the UI. They are called owners in the Resource Groups UI. I think we should use the same name for this construct here otherwise customers have to map between the concept of a resource group owner and this field name.

docs/resources/resource_group_project.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Required
+
+- `name` (String) The human-readable name of the resource. Values are case-sensitive.
+- `resource_group` (String)
+
+### Optional
+
+- `account_discovery` (Boolean)
+- `create_server_users` (Boolean) If `true`, 'sftd' (ASA Server Agent) creates corresponding local (unix or windows) user accounts in the ASA Project's servers.
```
waltergoulet-okta: Since this resource is unique to Okta PAM teams, I suggest you remove mention of ASA in those objects. I think it reads OK if you just drop it as a prefix for Projects and Server Agents (cc @isaacdurham-okta do you agree?)

docs/resources/resource_group_project.md
```diff
@@ -0,0 +1,46 @@
...
+- `forward_traffic` (Boolean) If `true`, all traffic in the ASA Project be forwarded through selected ASA Gateways.
+- `gateway_selector` (String) Assigns ASA Gateways with labels matching all selectors. At least one selector is required for traffic forwarding.
+- `next_unix_gid` (Number) The GID to use when creating a new ASA Server User. Default value starts at 63001.
```
waltergoulet-okta, line 27: Recognize that we don't expose these values in current OktaPA view of servers. Can we exclude them from the provider (next UID/GID)?

oktapam/constants/descriptions/attributes.go
```diff
@@ -49,6 +49,7 @@ var (
 	CreatedByUser                         = "The ASA User that created the resource."
 	CSRDetails                            = "Certificate Signing Request (CSR) Details."
 	DefaultAddress                        = "Default address of the gateway."
+	DelegatedAdminGroups                  = "The ids of the groups for users who are designated as delegated admins for the resource group"
```
waltergoulet-okta: Left comment in other file but repeating here; don't think we should call these delegated admin groups but owners instead to match UI. If must have this name, then add text to description calling out that it corresponds to the Owners field of a resource group.

oktapam/constants/descriptions/data_sources.go
```diff
@@ -16,17 +16,17 @@ var (
 	// Features In Progress
 
 	// Fetch
-	SourceADConnections = "A list of ASA AD Connections associated with an ASA Team."
-	SourceGateways      = fmt.Sprintf("Returns a list of all ASA Gateways connected to the ASA Team specified in the OKTAPAM_TEAM environment variable. %s", LinkGateway)
-	SourceGroup         = fmt.Sprintf("Returns a previously created ASA Group. %s", LinkGroup)
-	SourceProject       = fmt.Sprintf("Returns a previously created ASA Project. %s", LinkProject)
-	SourceProjectGroup  = fmt.Sprintf("Returns a previously created ASA Group assigned to a given ASA Project. %s", LinkProjectGroup)
-	SourceTeamSettings = fmt.Sprintf("Returns Team-level settings for a specific Team, such as authentication and enrollment details")
+	SourceADConnections          = "A list of ASA AD Connections associated with an ASA Team."
```
waltergoulet-okta: Mirroring earlier comment, if we remove ASA from the description text will the descriptions still work? Or if these descriptions contain ASA can we ensure that they are referenced only from objects that are ASA specific?

oktapam/resource_resource_group.go
```diff
@@ -0,0 +1,172 @@
...
+
+	id := d.Id()
+	resourceGroup, err := c.GetResourceGroup(ctx, id)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	resourceMap := resourceGroup.ToResourceMap()
+	for k, v := range resourceMap {
+		if k == attributes.ID {
```
sachinsaxena-okta, line 93: Any purpose to set the Id while reading it back ? Not sure, but there is some difference between setting the id attribute vs setting it as tf resource id.

oktapam/constants/descriptions/data_sources.go
```diff
@@ -16,17 +16,26 @@ var (
...
 	// Fetch
-	SourceADConnections          = "A list of ASA AD Connections associated with an ASA Team."
-	SourceGateways               = fmt.Sprintf("Returns a list of all ASA Gateways connected to the ASA Team specified in the OKTAPAM_TEAM environment variable. %s", LinkGateway)
-	SourceGroup                  = fmt.Sprintf("Returns a previously created ASA Group. %s", LinkGroup)
-	SourceProject                = fmt.Sprintf("Returns a previously created ASA Project. %s", LinkProject)
-	SourceProjectGroup           = fmt.Sprintf("Returns a previously created ASA Group assigned to a given ASA Project. %s", LinkProjectGroup)
-	SourceTeamSettings           = fmt.Sprintf("Returns Team-level settings for a specific Team, such as authentication and enrollment details")
-	SourceADUserSyncTaskSettings = fmt.Sprintf("Returns a previously created ASA AD User Sync Task Settings. %s", LinkADUserDiscovery)
+	SourceADConnections                      = "A list of ASA AD Connections associated with an ASA Team."
+	SourceGateways                           = fmt.Sprintf("Returns a list of all ASA Gateways connected to the ASA Team specified in the OKTAPAM_TEAM environment variable. %s", LinkGateway)
```
waltergoulet-okta, line 20: either omit ASA or change it to ASA/PAM

oktapam/constants/descriptions/data_sources.go
```diff
@@ -16,17 +16,26 @@ var (
...
+	SourceADUserSyncTaskSettings             = fmt.Sprintf("Returns a previously created ASA AD User Sync Task Settings. %s", LinkADUserDiscovery)
 
 	// List
-	SourceGroups                       = fmt.Sprintf("Returns a list of all ASA Groups associated with the ASA Team specified in the OKTAPAM_TEAM environment variable. %s", LinkGroup)
-	SourceProjects                     = fmt.Sprintf("Returns a list of ASA Projects associated with the ASA Team specified in the OKTAPAM_TEAM environment variable. %s", LinkProject)
-	SourceProjectGroups                = fmt.Sprintf("A list of ASA Project Groups associated with an ASA Project. %s", LinkProjectGroup)
-	SourceADUserSyncTaskSettingsIDList = fmt.Sprintf("Returns a list of previously created ASA AD User Sync Task Settings IDs for an AD connection. %s", LinkADUserDiscovery)
+	SourceADUserSyncTaskSettingsIDList        = fmt.Sprintf("Returns a list of previously created ASA AD User Sync Task Settings IDs for an AD connection. %s", LinkADUserDiscovery)
+	SourceGroups                              = fmt.Sprintf("Returns a list of all ASA Groups associated with the ASA Team specified in the OKTAPAM_TEAM environment variable. %s", LinkGroup)
```
waltergoulet-okta, line 34: Can we call this User Groups instead to be more descriptive?

oktapam/resource_resource_group_project.go
```diff
@@ -0,0 +1,266 @@
...
+		AccountDiscovery:   GetBoolPtrFromResource(attributes.AccountDiscovery, d, false),
+	}
+
+	resultingProject, err := c.CreateResourceGroupProject(ctx, project)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId(*resultingProject.ID)
+	// TODO: move this to be its own resource since it's lifecycle is outside of project
```
sachinsaxena-okta, line 144: Not very clear. So, Resource Group Project is already a separate resource.

oktapam/resource_resource_group_project_test.go
```diff
@@ -0,0 +1,210 @@
...
+	actualResourceGroupProject.ResourceGroupID = expectedResourceGroupProject.ResourceGroupID
+	actualResourceGroupProject.Team = expectedResourceGroupProject.Team
+
+	return nil
+}
+
+func testAccResourceGroupProjectCheckDestroy(resourceGroupName string, projectNames ...string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		client := testAccProvider.Meta().(client.OktaPAMClient)
+		resourceGroups, err := client.ListResourceGroups(context.Background())
```
sachinsaxena-okta, line 141: Not sure why we are checking for Resource Group instead of Resource Group Project ?

oktapam/constants/descriptions/data_sources.go
sachinsaxena-okta: I think better to remove ASA.

oktapam/resource_security_policy.go
```diff
@@ -0,0 +1,869 @@
...
+			attributes.ID: {
+				Type:     schema.TypeString,
+				Computed: true,
+				// Description is autogenerated
+			},
+			attributes.Name: {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: descriptions.Name,
+				ValidateDiagFunc: func(i any, p cty.Path) diag.Diagnostics {
```
sachinsaxena-okta, line 33: nit You can use the inbuilt validation functions too.

https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation

There is a method to convert legacy validate func - https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation#ToDiagFunc

docs/resources/resource_group.md
jamesatwill-okta: I get where Walter is coming from, and I agree if the API, backend, database columns, etc  also calls them owners. Otherwise I'd follow what the API calls them as the UI/UX can change, but we'll likely not seriously invest into renaming the API calls.

oktapam/resource_resource_group_project_test.go
SajanAlexander-okta: So each of these tests create a new resource group and should be removed.  Since the resource group deletion will cascade the deletion of a project, we can just check if the resource group is deleted.  

I ended up removing each of the resource-specific methods and just use the method for checking if a resource group is destroyed.

oktapam/resource_password_settings_test.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+}
+`
+
+func createTestAccPasswordSettingsCreateConfig(dgaName, resourceGroupName, projectName string) string {
+	return fmt.Sprintf(testAccPasswordSettingsCreateConfigFormat, dgaName, resourceGroupName, projectName)
+}
+
+const testAccPasswordSettingsUpdateConfigFormat = `
+resource "oktapam_group" "test_resource_group_dga_group" {
```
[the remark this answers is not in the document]
SajanAlexander-okta: As mentioned in another comment, I think that I'd rather do something like this in a separate PR that addresses formatting and possibly using the examples dir

oktapam/resource_password_settings_test.go
```diff
@@ -0,0 +1,264 @@
...
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: testAccProviders,
+		CheckDestroy:      testAccPasswordSettingsCheckDestroy(resourceGroupName, projectName),
+		Steps: []resource.TestStep{
+			{
+				Config: createTestAccPasswordSettingsCreateConfig(delegatedAdminGroupName, resourceGroupName, projectName),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					testAccPasswordSettingsCheckExists(resourceName, initialPasswordSettings),
```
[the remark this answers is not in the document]
sachinsaxena-okta: I think when resource is created we do a read and set those values in the state. So, if we are checking what we have in state then that should be enough. But, I see this is the pattern followed everywhere.

oktapam/constants/descriptions/attributes.go
```diff
@@ -8,87 +8,127 @@ import (
...
+	ADUserSyncTaskSettingsRunTest             = "If `true`, test is performed based on specified AD User Sync Task Settings"
+	ADUserSyncTaskSettingsSIDField            = "AD attribute mapped to user security identifier"
+	ADUserSyncTaskSettingsStartHourUTC        = "If AD user sync task is scheduled to run daily, then specify start hour in UTC"
+	ADUserSyncTaskSettingsUPNField            = "AD attribute mapped to user principal name"
+	AltNamesAttributes                        = "AD Attribute mapped to alternative hostnames or DNS entries used to resolve a discovered server."
+	ApproveDeviceWithoutInteraction           = "If enabled, ASA auto-approves devices for ASA Users that are authenticated into this Team."
+	BastionAttribute                          = "AD Attribute mapped to bastion host that Advanced Server Access clients can use to tunnel traffic to a discovered server."
+	CertificateCommonName                     = "Common Name or FQDN to which certificate is issued to."
+	CertificateContent                        = "Certificate Signing Request (CSR)/ Self Signed Certificate content."
```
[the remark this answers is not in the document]
SajanAlexander-okta: I'm not really familiar with this resource - @sachinsaxena-okta - could you answer?

oktapam/constants/descriptions/attributes.go
```diff
@@ -8,87 +8,127 @@ import (
...
+	CertificateID                             = "Certificate ID used for password less access method."
+	CertificateRequestType                    = "Specifies the type of certificate request - Certificate Signing Request (CSR)/ Self Signed Certificate."
```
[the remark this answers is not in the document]
SajanAlexander-okta: I'm not really familiar with this resource - @sachinsaxena-okta - could you answer?

oktapam/constants/descriptions/attributes.go
```diff
@@ -8,87 +8,127 @@ import (
...
+	CertificateStatus                         = "Certificate status - Valid/Request Created."
```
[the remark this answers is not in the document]
SajanAlexander-okta: I'm not really familiar with this resource - @sachinsaxena-okta - could you answer?

oktapam/constants/descriptions/attributes.go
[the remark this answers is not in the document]
sachinsaxena-okta: Possible Values: certificate_signing_request and self_signed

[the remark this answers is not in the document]
sachinsaxena-okta: Should we add it to the description like: Valid types are certificate_signing_request and self_signed.

oktapam/constants/descriptions/attributes.go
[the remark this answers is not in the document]
sachinsaxena-okta: @isaacdurham-okta Possible Values: valid and request_created

oktapam/constants/descriptions/attributes.go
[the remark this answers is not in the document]
sachinsaxena-okta: @isaacdurham-okta [AD Certificate Request Resource](https://registry.terraform.io/providers/okta/oktapam/latest/docs/resources/ad_certificate_request) is used to create certificate signing request/self signed certificate. This is the content returned after resource creation.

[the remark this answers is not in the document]
isaacdurham-okta: Maybe adjust wording to something like:
```suggestion
	CertificateContent                        = "The returned Certificate Signing Request (CSR) or Self Signed Certificate."
```

isaacdurham-okta (approved):

docs/resources/resource_group.md
SajanAlexander-okta: Based on conversations offline, we're going to keep the use of 'delegated resource admins' here and change the UX to match accordingly.

oktapam/resource_password_settings.go
```diff
@@ -0,0 +1,274 @@
...
+	passwordSettings, err := c.GetPasswordSettings(ctx, resourceGroupID, projectID)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	if passwordSettings != nil && len(passwordSettings.ManagedPrivilegedAccountsConfig) > 0 {
+		for k, v := range passwordSettings.ToResourceMap() {
+			if err := d.Set(k, v); err != nil {
+				diags = append(diags, diag.FromErr(err)...)
+			}
+		}
```
jamesbuchan-okta, line 159: Should this have set the Id here?

```
if diags == nil {
    d.SetId(formatPasswordSettingsID(resourceGroupID, projectID))
}
```

SajanAlexander-okta: Theoretically, it doesn't need to since a `Read` should already have an ID set.

PR #108, opened 2023-07-28
Add coarse grained attribute to oktapam tf provider.

leharbhatt-okta:
Modify "security_policy" resource in okta pam terraform provider to include "admin_level_permissions" 

Issue: https://oktainc.atlassian.net/browse/OKTA-630358


oktapam/resource_security_policy.go
```diff
@@ -787,20 +799,23 @@ func readPrivileges(privilegesAttr []any) ([]*client.SecurityPolicyRulePrivilege
 	return privileges, diags
 }
 
-func readPrivilegeEnabled(privilege any) (bool, bool) {
+func readPrivilegeEnabled(privilege any) (bool, bool, bool) {
```
sachinsaxena-okta, line 802: nit Better to just return (map[string] any, hasValue), otherwise require keep on adding new properties.

oktapam/resource_security_policy.go
```diff
@@ -220,6 +225,11 @@ func resourceSecurityPolicy() *schema.Resource {
 													Required:    true,
 													Description: descriptions.PrivilegeEnabled,
 												},
+												attributes.AdminLevelPermissions: {
```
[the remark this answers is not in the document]
sachinsaxena-okta: Actually, both can have different schema too in future. Ignore my above comment.

oktapam/client/security_policy.go
```diff
@@ -455,26 +455,30 @@ type SecurityPolicyRulePrivilege interface {
...
+	Enabled               *bool `json:"principal_account_rdp"`
+	AdminLevelPermissions *bool `json:"admin_level_permissions"`
 }
 
 func (*PrincipalAccountRDPPrivilege) isPrivilege() {}
 
 func (p *PrincipalAccountRDPPrivilege) ToResourceMap() map[string]any {
 	m := make(map[string]any, 1)
 	m[attributes.Enabled] = *p.Enabled
+	m[attributes.AdminLevelPermissions] = *p.AdminLevelPermissions
```
SajanAlexander-okta, line 467: can `p.AdminLevelPermissions` be nil?

oktapam/data_source_security_policies_test.go
```diff
@@ -107,9 +107,11 @@ resource "oktapam_security_policy" "test_ds_security_policies_1" {
 			}
 			principal_account_rdp {
```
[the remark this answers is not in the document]
SajanAlexander-okta: Yes - when a selector is used and broadly selects servers, regardless of the OS type.

oktapam/resource_security_policy.go
```diff
@@ -205,6 +205,11 @@ func resourceSecurityPolicy() *schema.Resource {
...
+													Type:        schema.TypeBool,
+													Required:    true,
```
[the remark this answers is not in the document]
SajanAlexander-okta: Yeah - I would set this to be optional and perhaps use a default of `false` here.

SajanAlexander-okta (changes requested):

oktapam/data_source_security_policies_test.go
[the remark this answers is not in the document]
sachinsaxena-okta: @SajanAlexander-okta Ok. From the UI, user first select "Session Type" which can be SSH or RDP. I think based on that either populate principal_account_ssh or principal_account_rdp. I am missing something here. May be it is just a UI constraint for now.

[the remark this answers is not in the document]
SajanAlexander-okta: Ah - yeah, you're right.  The UI has a restriction here where the API does not specifically have.

oktapam/resource_security_policy.go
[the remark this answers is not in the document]
leharbhatt-okta: I have now configured the API to return an error if Enabled = False and AdminLevelPermission=True

oktapam/resource_security_policy.go
```diff
@@ -741,7 +751,7 @@ func readPrivileges(privilegesAttr []any) ([]*client.SecurityPolicyRulePrivilege
 	privilegesM := privilegesAttr[0].(map[string]any)
 
 	if passwordCheckoutRDPI, ok := privilegesM[attributes.PasswordCheckoutRDP]; ok {
-		if enabled, hasValue := readPrivilegeEnabled(passwordCheckoutRDPI); hasValue {
+		if enabled, _, hasValue := readPrivilegeEnabled(passwordCheckoutRDPI); hasValue {
```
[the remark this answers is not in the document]
leharbhatt-okta: Added a new method to read admin-level access.

oktapam/data_source_security_policies_test.go
```diff
@@ -107,9 +107,11 @@ resource "oktapam_security_policy" "test_ds_security_policies_1" {
...
 				enabled = true
+				admin_level_permissions = true
 			}
 			principal_account_ssh {
 				enabled = true
+				admin_level_permissions = true
```
sachinsaxena-okta, line 114: Add at least one scenario where enabled = true and admin_level_permissions=false or admin_level_permissions is not even provided.

sachinsaxena-okta: I don't see any assertion done here for these properties - https://github.com/okta/terraform-provider-oktapam/blob/0f61daa7bae7f1056d06c4aeaab1b125a9867fcb/oktapam/data_source_security_policy_test.go#L35

leharbhatt-okta: added

sachinsaxena-okta (approved):

oktapam/resource_security_policy.go
```diff
@@ -845,3 +878,27 @@ func readPrincipalsFromResourceData(d *schema.ResourceData) (*client.SecurityPol
 
 	return principals, diags
 }
+
+func resourceSecurityPolicyCustomizeDiff(ctx context.Context, diff *schema.ResourceDiff, v interface{}) error {
+	return validatePrincipalAccounts(diff)
+}
+
+func validatePrincipalAccounts(diff *schema.ResourceDiff) error {
+	if diff.HasChange("rule.0.privileges.0.principal_account_ssh") {
```
SajanAlexander-okta, line 887: what happens when there are multiple rules?  does `rule.0` work?

oktapam/resource_security_policy.go
```diff
@@ -845,3 +878,33 @@ func readPrincipalsFromResourceData(d *schema.ResourceData) (*client.SecurityPol
 
 	return principals, diags
 }
+
+func resourceSecurityPolicyCustomizeDiff(ctx context.Context, diff *schema.ResourceDiff, v interface{}) error {
+	return validatePrincipalAccounts(diff)
+}
+
+func validatePrincipalAccounts(diff *schema.ResourceDiff) error {
```
SajanAlexander-okta, line 886: is the issue that you're trying to solve that the API won't accept a privilege that is not enabled but also has `admin_level_privileges` turned on?  or that it accepts it, but when the API returns the value, it doesn't return the `admin_level_privileges` value unless it is enabled?

SajanAlexander-okta (approved):
approved, but please respond to the comment i left

oktapam/resource_security_policy.go
leharbhatt-okta: So, API will not accept a privilege that is not enabled but has `admin_level_privileges` enabled. I added a check in Terraform,  so that we don't even make an API call if it is an invalid request.

PR #112, opened 2023-09-19
OKTA-643929 - secret folders resource

SajanAlexander-okta:
Adding resource for secret folders.  Note that this does not include the datasources yet, which will be covered in an additional PR.


oktapam/constants/descriptions/prefixes.go
```diff
@@ -4,4 +4,5 @@ const (
 	PrefixBeta = "Beta Feature:"
 	PrefixLEA  = "Limited Early Access Feature:"
 	PrefixDev  = "Feature under development:"
```
isaacdurham-okta, line 6: Small nit, but if we tweak this, it'll better match the other options.
```suggestion
	PrefixDev  = "Under Development Feature :"
```

oktapam/constants/descriptions/resources.go
```diff
@@ -32,5 +32,6 @@ var (
 	ResourceResourceGroup                      = fmt.Sprintf("%s A PAM construct that contains a group of projects. %s", PrefixLEA, LinkResourceGroup)
 	ResourceResourceGroupProject               = fmt.Sprintf("%s A PAM construct that contains a collection of resources that share settings. %s", PrefixLEA, LinkResourceGroupProject)
 	ResourceResourceGroupServerEnrollmentToken = fmt.Sprintf("%s A token used to enroll servers in a PAM Project. %s", PrefixLEA, LinkResourceGroupServerEnrollmentToken)
+	ResourceSecretFolder                       = fmt.Sprintf("%s A file-system like construct which can contain other secret folders or secrets.", PrefixGA)
```
isaacdurham-okta, line 35:
```suggestion
	ResourceSecretFolder                       = fmt.Sprintf("%s A file-system like construct which contains other secret folders or secrets.", PrefixGA)
```

isaacdurham-okta: Would something like this work instead?
```suggestion
	ResourceSecretFolder                       = fmt.Sprintf("%s A file-system like construct that contains secrets or nested secret folders.", PrefixGA)
```

oktapam/constants/descriptions/attributes.go
```diff
@@ -93,6 +93,7 @@ var (
 	NextUnixUID                               = "The UID to use when creating a new Server User. Default value starts at 60001."
 	OIDCIssuerURL                             = "The OIDC Issuer URL to use when configuring your Kubernetes cluster. "
 	OSAttribute                               = "The AD attribute that defines the operating system of discovered servers."
+	ParentFolder                              = "The secret folder which should contain this entity."
```
isaacdurham-okta, line 96:
```suggestion
	ParentFolder                              = "The secret folder which contains this entity."
```

isaacdurham-okta: Is the ParentFolder just the normal directory that contains the oktapam_secret_folder, or is there something special about it?
If it's just a normal directory, consider changing this string to:
```suggestion
	ParentFolder                              = "The directory which contains this entity."
```

isaacdurham-okta: Also, is "this entity" = the oktapam_secret_folder? If so, we might want to call that out a little more.

oktapam/client/secret_folder.go
```diff
@@ -0,0 +1,104 @@
+package client
+
+import (
```
sachinsaxena-okta, line 3: Not sure if we have openapi spec for secret folder apis. Can we add this to spec first and then generate sdk code ?

mariovillaplana-okta (commented):
Looks good to me, but we don't want to merge this until closer to feature release, right? How are we ensuring that unreleased features aren't making their way into usable constructs?

PR #113, opened 2023-10-20
create both sdk and local client

sachinsaxena-okta:
Integrate sdk client on provider initialization.


.github/workflows/release.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.18
+          go-version: 1.20
```
sachinsaxena-okta, line 32: Need to upgrade this because sdk use url.JoinPath which is in newer version

jamesatwill-okta: I think you could bump this to 1.21. Does it use the latest patch version automatically?

oktapam/client/okta_pam_client.go
```diff
@@ -196,3 +198,35 @@ func stringSliceToInterfaceSlice(arr []string) []any {
...
+func CreateSDKClient(providerConfig *OktaPAMProviderConfig) (*pam.APIClient, error) {
+	apiClientConfigOpts := []pam.ConfigOption{
+		pam.WithHost(providerConfig.APIHost),
+		pam.WithTeam(providerConfig.Team),
+		pam.WithAPIKey(providerConfig.APIKey),
+		pam.WithAPISecret(providerConfig.APIKeySecret),
+		pam.WithUserAgent(terraformUserAgent),
+	}
+
+	override, present := os.LookupEnv(TRUSTED_DOMAIN_OVERRIDE_ENV_VAR)
```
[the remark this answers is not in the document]
jamesatwill-okta: TRUSTED_DOMAIN_OVERRIDE is really just for exceptional cases. I think I'd rather it remain an env var, but could be convinced otherwise?

.github/workflows/release.yml
sachinsaxena-okta: I think it apply the latest patch version

oktapam/provider.go
```diff
@@ -137,15 +138,33 @@ func Provider() *schema.Provider {
...
+		SDKClient:   sdkClient,
+		LocalClient: localClient,
+	}, nil
 }
+
+func getSDKClientFromMetadata(meta interface{}) *pam.APIClient {
+	return meta.(*client.APIClients).SDKClient
+}
+
+func getLocalClientFromMetadata(meta interface{}) *client.OktaPAMClient {
```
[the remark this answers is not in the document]
sachinsaxena-okta: Fixed

.github/workflows/release.yml
sachinsaxena-okta: Fixed

jamesatwill-okta (approved):

Makefile.ci
```diff
@@ -57,7 +57,7 @@ ci-check-generate: ci-container
 		-e BUILDKITE_BUILD_NUMBER \
 		-e BUILDKITE_PIPELINE_SLUG \
 		-e BUILDKITE_JOB_ID \
-		-e TF_ACC_TERRAFORM_PATH=/usr/local/bin/terraform \
```
sachinsaxena-okta, line 60: Not needed as it's already set as part of ci-check-generated.sh

PR #115, opened 2023-11-09
OKTA-649266 - update security policy for GA features

SajanAlexander-okta:
Updates security policies to allow for GA features

 - secrets / secret folders
 - MFA condition
 - constrain security policy for resource group


oktapam/constants/descriptions/attributes.go
```diff
@@ -89,6 +90,8 @@ var (
 	LocalAccount                              = "The username of a local account on a server."
 	LocalAccounts                             = "The usernames of a local accounts on a server."
 	ManagedPrivilegedAccounts                 = "An array of managed accounts for password rotation."
+	MFAACRValues                              = "The authentication context class reference value used to defines a specific set of assurance level requirements that the protected resource requires."
```
SajanAlexander-okta, line 93: @isaacdurham-okta - could you take a look at these descriptions?

oktapam/resource_security_policy.go
```diff
@@ -43,6 +42,11 @@ func resourceSecurityPolicy() *schema.Resource {
 				Required:    true,
 				Description: descriptions.SecurityPolicyActive,
 			},
+			attributes.ResourceGroup: {
```
sachinsaxena-okta, line 45: Should this be ResourceGroupID ?

oktapam/resource_security_policy.go
```diff
@@ -83,9 +87,48 @@ func resourceSecurityPolicy() *schema.Resource {
...
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
+									attributes.Secrets: {
+										Type:        schema.TypeList,
+										Optional:    true,
+										MinItems:    1,
+										MaxItems:    1,
+										Description: descriptions.SecurityPolicySecrets,
+										Elem: &schema.Resource{
+											Schema: map[string]*schema.Schema{
```
sachinsaxena-okta, line 97: It's fine to add all the schema in the parent SecurityPolicy schema, but for readability I like creating separate schema type and then referencing it.

oktapam/resource_security_policy.go
```diff
@@ -83,9 +87,48 @@ func resourceSecurityPolicy() *schema.Resource {
...
+												attributes.Secret: {
```
sachinsaxena-okta, line 98: I think only Secret or SecretFolder can be specified but not both. You can use ConflictsWith property in the schema. Take a look at - https://github.com/okta/terraform-provider-oktapam/blob/8505915c86f1eb7fb65cfcf19b826fed65f55516/oktapam/resource_ad_certificate_request.go#L72

oktapam/resource_security_policy.go
```diff
@@ -83,9 +87,48 @@ func resourceSecurityPolicy() *schema.Resource {
...
+													Elem: &schema.Resource{
+														Schema: map[string]*schema.Schema{
+															attributes.SecretID: {
+																Type:        schema.TypeString,
+																Required:    true,
+																Description: descriptions.SecretID,
+															},
+														},
+													},
+												},
```
sachinsaxena-okta, line 111: Can policy have multiple secret folders ?

oktapam/resource_security_policy.go
```diff
@@ -164,6 +207,56 @@ func resourceSecurityPolicy() *schema.Resource {
...
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
+									attributes.Secret: {
+										Type:        schema.TypeList,
+										Optional:    true,
+										MaxItems:    1,
+										Description: descriptions.PrivilegeSecret,
+										Elem: &schema.Resource{
+											Schema: map[string]*schema.Schema{
+												attributes.List: {
```
sachinsaxena-okta, line 217: Why these privileges need to be required ? Also, there are different privileges for secret and folder. If in the policy, we have a SecretFolder resource then what is the relevance of PrivilegeSecret*.

oktapam/resource_security_policy.go
```diff
@@ -386,13 +498,26 @@ func readPolicyFromResourceData(d *schema.ResourceData) (client.SecurityPolicy,
...
+
 	policy := client.SecurityPolicy{
 		ID:          id,
 		Name:        GetStringPtrFromResource(attributes.Name, d, false),
 		Active:      GetBoolPtrFromResource(attributes.Active, d, false),
 		Description: GetStringPtrFromResource(attributes.Description, d, false),
 	}
 
+	if resourceGroupID := GetStringPtrFromResource(attributes.ResourceGroup, d, false); resourceGroupID != nil {
+		if MatchesUUID(*resourceGroupID) {
```
sachinsaxena-okta, line 510: I think UUID check can be part of the schema validation.

oktapam/constants/descriptions/attributes.go
```diff
@@ -54,6 +54,7 @@ var (
 	ConditionAccessRequestRequestTypeID       = "The ID of the Access Request."
 	ConditionAccessRequestRequestTypeName     = "The name of the Access Request."
 	ConditionGateway                          = "Configures traffic settings for an existing Gateway."
+	ConditionMFA                              = "Configures Multi-Factor Auth settings required to access the resource."
```
isaacdurham-okta, line 57:
```suggestion
	ConditionMFA                              = "Configures multifactor auth settings required to access the resource."
```

oktapam/constants/descriptions/attributes.go
```diff
@@ -113,11 +125,17 @@ var (
 	RequirePreauthForCreds                    = "If `true`, requires preauthorization before a User can retrieve credentials to sign in."
 	ResourceGroupID                           = "The UUID of a PAM Resource Group."
 	Roles                                     = "A list of roles for the ASA Group. Options are 'access_user', 'access_admin', and 'reporting_user'."
+	SecretFolderID                            = "The UUID of the secret folder."
```
isaacdurham-okta, line 128:
```suggestion
	SecretFolderID                            = "The UUID of the Secret Folder."
```

oktapam/constants/descriptions/attributes.go
```diff
@@ -113,11 +125,17 @@ var (
...
+	SecretID                                  = "The UUID of the secret."
```
isaacdurham-okta, line 129:
```suggestion
	SecretID                                  = "The UUID of the Secret."
```

oktapam/constants/descriptions/attributes.go
```diff
@@ -113,11 +125,17 @@ var (
...
 	SecurityPolicyActive                      = "If true, indicates that the Security Policy is active."
 	SecurityPolicyLabelSelectors              = "Defines the label selectors used to target resources by the Security Policy."
 	SecurityPolicyPrincipals                  = "Defines the users bound to the Security Policy."
+	SecurityPolicyResouceGroup                = "The ID of the Resource Group to which the Security Policy is constrained.  If this is not set, the Security Policy will apply to all resources for the team.  This value must be set if the user under which the terraform is being applied is a Delegated Security Admin."
```
isaacdurham-okta, line 133:
```suggestion
	SecurityPolicyResouceGroup                = "The UUID of a specific Resource Group associated with the Security Policy. If undefined, the Security Policy applies to all resources for your Team. This value must be defined if the current user has the Delegated Security Admin role."
```

oktapam/constants/descriptions/attributes.go
```diff
@@ -113,11 +125,17 @@ var (
...
 	SecurityPolicyResources                   = "Defines the resources targeted by the Security Policy."
 	SecurityPolicyRule                        = "Defines the privileges available to resources matched to the Security Policy."
+	SecurityPolicySecret                      = "Defines a specific secret targeted by the Security Policy."
+	SecurityPolicySecretFolder                = "Defines a specific secret folder targeted by the Security Policy."
```
isaacdurham-okta, line 137:
```suggestion
	SecurityPolicySecretFolder                = "Defines a specific Secret Folder targeted by the Security Policy."
```

oktapam/constants/descriptions/attributes.go
```diff
@@ -113,11 +125,17 @@ var (
...
+	SecretFolderID                            = "The UUID of the secret folder."
+	SecretID                                  = "The UUID of the secret."
 	SecurityPolicyActive                      = "If true, indicates that the Security Policy is active."
 	SecurityPolicyLabelSelectors              = "Defines the label selectors used to target resources by the Security Policy."
 	SecurityPolicyPrincipals                  = "Defines the users bound to the Security Policy."
+	SecurityPolicyResouceGroup                = "The ID of the Resource Group to which the Security Policy is constrained.  If this is not set, the Security Policy will apply to all resources for the team.  This value must be set if the user under which the terraform is being applied is a Delegated Security Admin."
 	SecurityPolicyResources                   = "Defines the resources targeted by the Security Policy."
 	SecurityPolicyRule                        = "Defines the privileges available to resources matched to the Security Policy."
+	SecurityPolicySecret                      = "Defines a specific secret targeted by the Security Policy."
```
isaacdurham-okta, line 136:
```suggestion
	SecurityPolicySecret                      = "Defines a specific Secret targeted by the Security Policy."
```

oktapam/constants/descriptions/attributes.go
```diff
@@ -89,6 +90,8 @@ var (
 	LocalAccount                              = "The username of a local account on a server."
 	LocalAccounts                             = "The usernames of a local accounts on a server."
 	ManagedPrivilegedAccounts                 = "An array of managed accounts for password rotation."
+	MFAACRValues                              = "The authentication context class reference value used to defines a specific set of assurance level requirements that the protected resource requires."
+	MFAReAuthFrequencyInSeconds               = "The number of seconds a successful MFA is valid for before requiring another MFA to access the resource again."
```
isaacdurham-okta, line 94: Matching this to the spec description. We could maybe trim it more if we wanted.
```suggestion
	MFAReAuthFrequencyInSeconds               = "The number of seconds an MFA verification remains valid. After this time users need to reauthenticate before they can open new connections to resources. Defining a value of `0` requires users to authenticate for every connection."
```

oktapam/constants/descriptions/attributes.go
isaacdurham-okta:
```suggestion
	MFAACRValues                              = "The authentication context class reference (ACR) for this policy. This defines a specific set of assurance level requirements required by a protected resource."
```

isaacdurham-okta (approved):

docs/data-sources/security_policy.md
```diff
@@ -69,6 +70,15 @@ Read-Only:
 - `traffic_forwarding` (Boolean)
 
 
+<a id="nestedobjatt--rule--conditions--mfa"></a>
+### Nested Schema for `rule.conditions.mfa`
```
mariovillaplana-okta, line 74: It's hard for me to tell - does this allow them to try to set up MFA conditions on secret rules? That's not currently supported. I suppose the same question would apply to gateway conditions.

oktapam/resource_security_policy.go
mariovillaplana-okta: No, I think it's a named object so this should work fine

oktapam/resource_security_policy.go
mariovillaplana-okta: A policy, yes, but not in a single rule

mariovillaplana-okta (approved):
Approving since the API will surface the error if they try to add an MFA condition to a secret rule, but just wondering about the current behavior. Please test locally before merging

oktapam/resource_security_policy.go
sachinsaxena-okta: @mariovillaplana-okta I didn't get it. We pass ResourceGroupID for this attribute, correct ?

docs/data-sources/security_policy.md
SajanAlexander-okta: We constrain gateways, but not MFA - the main reason is that while this is not currently supported, we don't want to have to roll another version of the TF provider and migrate people when support is added.

oktapam/resource_security_policy.go
SajanAlexander-okta: From what I've seen, we generally use the term without the `ID` as attribute names, but have the description indicate that the expectation is the UUID

oktapam/resource_security_policy.go
SajanAlexander-okta: Because we use TypeLists/TypeSets, using ConflictsWith is kind of a nightmare.  I ended up doing the validation within the code itself.

oktapam/resource_security_policy.go
SajanAlexander-okta: I've constrained this.

oktapam/resource_security_policy.go
SajanAlexander-okta: My goal here is to be explicit on the privileges granted, even if the default is false.  

There are not different privileges for a secret/folder.

PR #122, opened 2024-01-22
Add Database Resource & Datasource

jamesbuchan-okta:
Initial implementation of databases.

One tricky thing is that we take in a plain text password, but then translate that to an EncryptedString before making the api request.  Neither the plain text password or the encryped password is returned in the GET api request, so we must manually store it in state ourselves.

Interfaces were made to try and make this as generic as possible, as the use-case will likely reappear for other resources in future.

It is currently handled by this block.
```
wrap := wrappers.DatabaseResourceResponseWrapper{*database}
attributeOverrides := utils.GenerateAttributeOverrides(d, wrap)
for key, value := range wrap.ToResourceMap(attributeOverrides) {
	if err := d.Set(key, value); err != nil {
		return diag.FromErr(err)
	}
}
```


vendor/modules.txt
```diff
@@ -36,7 +36,7 @@ github.com/apparentlymart/go-textseg/v15/textseg
 # github.com/armon/go-radix v0.0.0-20180808171621-7fddfc383310
 ## explicit
 github.com/armon/go-radix
-# github.com/atko-pam/pam-sdk-go v1.0.53
+# github.com/atko-pam/pam-sdk-go v1.0.54-0.20240122191100-fa65b5f2cf4d
```
jamesbuchan-okta, line 39: Need to update this once the sdk PR is merged.

oktapam/resource_database.go
```diff
@@ -0,0 +1,418 @@
...
+				Description: descriptions.RecipeBookID,
+			},
+			attributes.ManagementConnectionDetails: {
+				Type:        schema.TypeList,
+				Required:    true,
+				MinItems:    1,
+				MaxItems:    1,
+				Description: descriptions.ManagementConnectionDetails,
+				Elem:        managementConnectionDetails,
+				ForceNew:    false,
```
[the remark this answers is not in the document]
jamesbuchan-okta: I think I had this true when I first started this.  Do you know if there are any that should be true?

Looking over the platform code, I think `DatabaseType` and `ManagementConnectionDetailsType` are the only two?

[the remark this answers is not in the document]
sachinsaxena-okta: Yes that's true.

oktapam/resource_database.go
```diff
@@ -0,0 +1,418 @@
...
+			},
+			attributes.ManagementConnectionDetailsType: {
```
SajanAlexander-okta, line 111: With security policies, I opted to use a different method for cases where you have a polymorphic type.  With this, it would end up being something like: 

```
management_connection_details {
  mysql {
   ....
  }
}
```

It does make the code more complicated, but I like it since it hides some of the API implementation details from the user and reduces chances for misconfiguration.

jamesbuchan-okta: Yeah, I like that.

I was also curious if we'd want to remove some of the nesting of auth details & connection details and have them all under `management_connection_details`.  Is that what you're thinking?

e.g.

```
management_connection_details {
  mysql {
     host: ...
     port: ...
     user: ...
     pass: ...
  }
}
```

vs. separating out username, password, secret?

SajanAlexander-okta: i believe so - the main thing is if all of that stuff is contained within the polymorphic object the API expects, then it should be under `management_connection_details.mysql`.  if it is something that's common to all API requests, regardless of the details type, then it should remain outside of it.

jamesbuchan-okta: noting for historical purposes, we'll be going with something like

```
management_connection_details {
  mysql {
     host: ...
     port: ...
     basic_auth {
          user: ...
          pass: ...
     }
  }
}
```

oktapam/resource_database.go
```diff
@@ -0,0 +1,420 @@
...
+
+	d.SetId(database.Id)
+
+	wrap := wrappers.DatabaseResourceResponseWrapper{*database}
+	attributeOverrides := utils.GenerateAttributeOverrides(d, wrap)
+	for key, value := range wrap.ToResourceMap(attributeOverrides) {
+		if err := d.Set(key, value); err != nil {
+			return diag.FromErr(err)
+		}
+	}
```
jamesbuchan-okta, line 214: This was done to make the setting of the password attribute more generic, and hopefully easier to plug in overrides for other resources when needed in the future.

oktapam/resource_database.go
```diff
@@ -0,0 +1,420 @@
...
+func parseDatabaseID(resourceId string) (string, string, string, error) {
+	split := strings.Split(resourceId, "/")
+	if len(split) != 3 {
+		return "", "", "", fmt.Errorf("expected format: <resource_group_id>/<project_id>/<id>, received: %s", resourceId)
+	}
+	return split[0], split[1], split[2], nil
+}
+
+func mgmtConnectionDetailsFromResource(_ context.Context, pamClient client.SDKClientWrapper, d *schema.ResourceData) (string, *pam.ManagementConnectionDetails, diag.Diagnostics) {
+	v, ok := d.GetOk(attributes.ManagementConnectionDetails)
```
sachinsaxena-okta, line 320: You don't need these checks. It's already part of schema.

oktapam/resource_database.go
```diff
@@ -0,0 +1,420 @@
...
+
+func resourceDatabaseCreate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	c := getSDKClientFromMetadata(m)
+	resourceGroupID := d.Get(attributes.ResourceGroup).(string)
+	projectID := d.Get(attributes.Project).(string)
+	canonicalName := d.Get(attributes.CanonicalName).(string)
+	dbType := d.Get(attributes.DatabaseType).(string)
+
+	var selectorLabels *map[string]string
+	if selector, ok := d.GetOk(attributes.ManagementGatewaySelector); ok {
```
sachinsaxena-okta, line 165: nit This logic we can pull out as utility method to convert tf map to go map[string]string.

oktapam/resource_database.go
jamesbuchan-okta: i was starting to realize we can trust the schema for more of this.  will have another scan through and clean things up

oktapam/provider.go
```diff
@@ -115,6 +117,7 @@ func Provider() *schema.Provider {
 		DataSourcesMap: map[string]*schema.Resource{
 			providerADConnectionsKey:                       dataSourceADConnections(),
 			providerCurrentUser:                            dataSourceCurrentUser(),
+			providerDatabaseKey:                            dataSourceDatabase(),
```
SajanAlexander-okta, line 120: generally we have a list-based datasource as well.  the usage is really more around obtaining the id for a resource based on a filter, so you may do something like: 

```
data oktapam_databases db-list {
  resource_group = '...'
  project = '...'
  name = 'YourMySQL'
}

data oktapam_database yourmysql {
  id = data.oktapam_databases.db-list[0].id
}
```

will we be supporting that?

sachinsaxena-okta: We will be supporting that. We need to add API for listing all database resources first.

SajanAlexander-okta: That makes sense, thanks @sachinsaxena-okta

oktapam/resource_database.go
```diff
@@ -0,0 +1,418 @@
...
+
+	overrides := make(map[string]any, 1)
+	if newPassword != nil && *newPassword != "" {
+		overrides[attributes.Password] = *newPassword
+	} else if password, ok := d.Get(attributes.NestedManagementConnectionPassword).(string); ok {
+		// if a refresh happens the password may exist in state
+		overrides[attributes.Password] = password
+	}
+
+	d.SetId(database.Id)
```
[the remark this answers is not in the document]
jamesbuchan-okta: Removed in 3608574a504bb4213ac00b172682c23fe98226dc

oktapam/resource_database.go
jamesbuchan-okta: Removed in 3608574a504bb4213ac00b172682c23fe98226dc

oktapam/resource_database.go
jamesbuchan-okta: Moved this into a util func, `GetMapPtrFromResource`

oktapam/resource_database.go
```diff
@@ -0,0 +1,380 @@
...
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: descriptions.ManagementGatewaySelectorID,
+			},
+			attributes.CreatedAt: {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: descriptions.CreatedAt,
+			},
+			attributes.UpdatedAt: {
```
SajanAlexander-okta, line 145: I don't think we generally expose this within the provider.  Not totally opposed to keeping it, but I don't think it's really that useful.

oktapam/client/wrappers/database.go
```diff
@@ -0,0 +1,139 @@
...
+	m[attributes.ManagementConnectionDetails] = mgmtDetails
+
+	return m
+}
+
+func (w DatabaseResourceResponseWrapper) AttributeOverridePaths() []string {
+	overrides := []string{}
+	childOverrides := ManagementConnectionDetailsWrapper{w.ManagementConnectionDetails}.AttributeOverridePaths()
+	for _, attr := range childOverrides {
+		overrides = append(overrides, fmt.Sprintf("%s.0.%s", attributes.ManagementConnectionDetails, attr))
```
SajanAlexander-okta, line 68: same comment for others below, but it would be good to have comments on what each part of the path means.  also, is the '0' value/index guaranteed?

docs/resources/database.md
```diff
@@ -0,0 +1,69 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "oktapam_database Resource - terraform-provider-oktapam"
```
SajanAlexander-okta, line 3: We will want to use a template for resources with sensitive information that will be stored within the state file.  See https://github.com/hashicorp/terraform-provider-tls/blob/main/templates/resources/private_key.md.tmpl for an example of this and https://github.com/hashicorp/terraform-plugin-docs?tab=readme-ov-file#how-it-works for an explanation of the docs plugin.  

Also note that we're on a pretty old version of the docs plugin, so we may need to update it if we run into any issues with templating.

oktapam/utils/utils.go
```diff
@@ -113,3 +115,32 @@ func CreateCheckResourceDestroy(typeName string, checkResourceExists checkResour
 		return nil
 	}
 }
+
+// GenerateAttributeOverrides will evaluate which attributes on the resource may have overrides set,
```
SajanAlexander-okta, line 119: could you update the comment to include reasons why you would want to use this?

oktapam/resource_database.go
SajanAlexander-okta: Could you explain what the deal was with password that required this?

SajanAlexander-okta (changes requested):

oktapam/data_source_database.go
```diff
@@ -0,0 +1,108 @@
...
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+			attributes.ManagementGatewaySelectorID: {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: descriptions.ManagementGatewaySelectorID,
+			},
+			attributes.CreatedAt: {
```
sachinsaxena-okta, line 69: Usually we are not keeping create_at/update_at fields in datasource or resources as Sajan mentioned in other comment. Also, don't think you are setting these while converting ToResourceMap.

oktapam/param_utils.go
```diff
@@ -134,6 +134,24 @@ func GetTypeListMapFromResourceElement(attr string, data map[string]any) map[str
 	return listArr[0].(map[string]any)
 }
 
+func GetMapPtrFromResource[V any](attr string, d *schema.ResourceData) (*map[string]V, diag.Diagnostics) {
```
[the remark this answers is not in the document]
jamesbuchan-okta: Good question... I see ManagementGatewaySelector is optional in the spec so it is a pointer in the sdk.

```
ManagementGatewaySelector       *map[string]string              `json:"management_gateway_selector,omitempty"`
```

Looking at the platform code we check whether it is set or not. e.g. from the create handler.

```
if apiDBResourceRequest.ManagementGatewaySelector != nil && len(apiDBResourceRequest.ManagementGatewaySelector) > 0 {
    resGatewaySelector.Init(team.ID)
    resGatewaySelector.Selector = apiDBResourceRequest.ManagementGatewaySelector
    if err := tx.CreateResourceGatewaySelector(ctx, team.ID, resGatewaySelector); err != nil {
        return err
    }
    dbResource.ManagementGatewaySelectorID = utils.NewNullString(resGatewaySelector.ID)
}
```

oktapam/resource_database.go
```diff
@@ -0,0 +1,380 @@
...
+		})
+	}
+	if projectID == "" {
+		diags = append(diags, diag.Diagnostic{
+			Severity: diag.Error,
+			Summary:  "could not obtain project id from resource",
+		})
+	}
+
+	if diags != nil {
```
sachinsaxena-okta, line 246: You don't need to check all the above error conditions related to dbID, resourceGroupID, projectID. These attributes are mandatory in the resource schema, so while updating the same resource these must be present. I should have put this comment in the earlier feedback.

oktapam/resource_database.go
jamesbuchan-okta: This made me realize I wasn't actually setting them, nor `RecipeBook` - forgot those top level fields after focussing on the other problems.  I've removed those and now actually set the recipe.

oktapam/resource_database.go
jamesbuchan-okta: and im not sure why i didnt remove them last night.

oktapam/utils/utils.go
jamesbuchan-okta: I'm hoping [this](https://github.com/okta/terraform-provider-oktapam/pull/122/commits/990049a5bfe8ca59562ab23b24ad44d26779f7b5#diff-50ff6165be9dcbc6b7ac48292623314829fbdf124fba1537f184598594d087a9R120) may take care of all your questions around this.

I very much welcome feedback if you think there is a better way to deal with this.  I couldn't find much while searching about this kinda problem.

oktapam/data_source_database.go
jamesbuchan-okta: removed

sachinsaxena-okta (dismissed):
LGTM

oktapam/client/wrappers/database.go
jamesbuchan-okta: done... hopefully it's sufficient.

docs/resources/database.md
jamesbuchan-okta: Added a template but not entirely sure what all we want to say.  You probably saw the conversation a while back where I was trying to understand the best practice for people using this and how they'd provide a password... but I never quite got to the full answer.

jamesbuchan-okta: I also populated the template to be as similar to the auto generated one as possible (plus the addition of the Security Notice section)

templates/resources/database.md.tmpl
```diff
@@ -0,0 +1,14 @@
...
+description: |-
+  {{ .Description }}
+---
+
+# {{ .Name }} ({{ .Type }})
+
+{{ .Description }}
+
+## Security Notice
+The password specified in this resource will be stored *unencrypted* in your Terraform state file.
```
SajanAlexander-okta, line 12:
```suggestion
The password specified in this resource will be stored *unencrypted* in your Terraform state and plan files.  Please take precautions to store these in a secure location.
```

oktapam/utils/utils.go
SajanAlexander-okta: Makes sense.  Thanks

oktapam/param_utils.go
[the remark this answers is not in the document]
SajanAlexander-okta: Ok - it's unnecessary, but it seems like the auto-generation from the SDK doesn't take it into account

SajanAlexander-okta (changes requested):

SajanAlexander-okta (approved):

PR #124, opened 2024-01-23
Add database password settings Resource & Datasource

jamesbuchan-okta:
This is largely a copy of the existing password settings for servers but uses the SDK; cannot use the `ManagedPrivilegedAccountsConfig` field; and has to handle "deleted" slightly different.


oktapam/resource_database.go
```diff
@@ -0,0 +1,380 @@
+package oktapam
```
SajanAlexander-okta, line 1: should this file be in the PR?  it seems like its real definition is within #122

jamesbuchan-okta: Sorry, I pinged Sachin about this PR but not you.  This is rebased on 122 since it uses some of the same changes.  I figured it could be merged first then this one should be easy and hopefully not need another rebase.

oktapam/data_source_database_password_settings.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	if passwordPolicy != nil {
+		wrap := wrappers.PasswordPolicyWrapper{PasswordPolicy: *passwordPolicy, UseManagedPrivilegedAccountsConfig: false}
+		for k, v := range wrap.ToResourceMap(nil) {
+			if err := d.Set(k, v); err != nil {
+				diags = append(diags, diag.FromErr(err)...)
+			}
+		}
+		if diags == nil {
+			d.SetId(formatPasswordSettingsID(resourceGroupID, projectID))
```
sachinsaxena-okta, line 87: Id for server password setting and db password setting will be same in this case and conflict.

sachinsaxena-okta: Ignore this comment. My bad. It's different resource so same id is fine.

SajanAlexander-okta (approved):

oktapam/data_source_database_password_settings.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/okta/terraform-provider-oktapam/oktapam/client"
+	"github.com/okta/terraform-provider-oktapam/oktapam/client/wrappers"
+	"github.com/okta/terraform-provider-oktapam/oktapam/constants/attributes"
+	"github.com/okta/terraform-provider-oktapam/oktapam/constants/descriptions"
+)
+
+func dataSourceDatabasePasswordSettings() *schema.Resource {
+	return &schema.Resource{
+		Description: descriptions.SourcePasswordSettings,
```
sachinsaxena-okta, line 16: This should be SourceDatabasePasswordSettings

oktapam/resource_database_password_settings.go
```diff
@@ -0,0 +1,258 @@
...
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Description: descriptions.CharacterOptionsRequireFromEachSet,
+		},
+	},
+}
+
+func resourceDatabasePasswordSettings() *schema.Resource {
+	return &schema.Resource{
+		Description:   descriptions.ResourcePasswordSettings,
```
sachinsaxena-okta, line 49: This also should be separate one for databasepasswordsettings

oktapam/data_source_database_password_settings.go
jamesbuchan-okta: Pushed a commit to add the new one and update the existing to specify it's for servers.

sachinsaxena-okta (approved):

PR #125, opened 2024-01-31
Combine SDKV2 and Framework Plugin Provider

sachinsaxena-okta:
This PR is to support adding resources built using new TF Plugin Framework.

SDKV2 used for tf plugin development is designed for maintaining tf plugins that are compatible with Plugin Protocol version 5. Plugins need to communicate with Terraform CLI, protocol version 5 is supported by CLI version 0.12 and later. Protocol version 6 support tf cli version 1.0 or later.

To start using new TF Plugin Framework(https://developer.hashicorp.com/terraform/plugin/framework) we have two options:

1. Downgrade new plugin framework server to support protocol version 5
2. Upgrade old SDKV2 provider server to support protocol version 6

If we go with Option 1, then will not be able to use some of the newer features like [Nested Attributes](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/attributes#nested-attribute-types) 

Going with option 2, will require upgrading tf cli version to 1.0+. I am going with option 2.


oktapam/fwprovider/framework_provider.go
sachinsaxena-okta, line 1: I think we should follow this structure -

oktapam
- provider
-- framework
--- provider.go
-- sdkv2
--- provider.go
-- factory.go (This is to create a provider Server Factory. This code currently we are duplicating in provider_test.go and main.go)

sachinsaxena-okta: Resources and Resource Test can be at the same place for all the providers. So, basically keep them under - oktapam/. I also observed that you havn't created datasource for this.

oktapam/fwprovider/resource_server_checkout_settings.go
sachinsaxena-okta, line 1: Move this file under oktapam/

oktapam/fwprovider/resource_server_checkout_settings.go
```diff
@@ -0,0 +1,277 @@
...
+			"id": schema.StringAttribute{
+				Computed: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"checkout_duration_in_seconds": schema.Int64Attribute{
+				Optional:            true,
+				Computed:            true,
+				Description:         "The duration in seconds for the checkout. If the checkout is enabled, the duration is the maximum time a user can access the resource before the checkout expires.",
```
sachinsaxena-okta, line 62: We are keeping description in oktapam/constants/descriptions.

oktapam/fwprovider/resource_server_checkout_settings.go
```diff
@@ -0,0 +1,277 @@
...
+				MarkdownDescription: "If provided, only the account identifiers listed are excluded from the checkout requirement. This list is only considered if `checkout_required` is set to `true`. Only one of `include_list` and `exclude_list` can be specified in a request since they are mutually exclusive.",
+			},
+			"include_list": schema.ListAttribute{
+				ElementType:         types.StringType,
+				Optional:            true,
+				Description:         "If provided, only the account identifiers listed are required to perform a checkout to access the resource. This list is only considered if `checkout_required` is set to `true`. Only one of `include_list` and `exclude_list` can be specified in a request since they are mutually exclusive.",
+				MarkdownDescription: "If provided, only the account identifiers listed are required to perform a checkout to access the resource. This list is only considered if `checkout_required` is set to `true`. Only one of `include_list` and `exclude_list` can be specified in a request since they are mutually exclusive.",
+			},
+			"project": schema.StringAttribute{
```
sachinsaxena-okta, line 85: This should have force new true. If project is changed in the resource that means deleting the setting for old project and creating new one.

oktapam/fwprovider/resource_server_checkout_settings.go
```diff
@@ -0,0 +1,277 @@
...
+				Required:            true,
+				Description:         "The UUID of a Project",
+				MarkdownDescription: "The UUID of a Project",
+			},
+			"resource_group": schema.StringAttribute{
```
sachinsaxena-okta, line 90: This should have force new 'true'. If resource group is changed in the resource that means deleting the setting for old project and creating new one.

oktapam/fwprovider/resource_server_checkout_settings.go
```diff
@@ -0,0 +1,277 @@
...
+
+	if err != nil {
+		resp.Diagnostics.AddError("Error creating server checkout settings", err.Error())
+		return
+	}
+	// Set state to fully populated data
+	plan.Id = types.StringValue(formatServerCheckoutSettingsID(plan.ResourceGroup, plan.Project))
+	diags = resp.State.Set(ctx, plan)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
```
sachinsaxena-okta, line 126: What is the purpose of doing this? Errors are already added in resp.Diagnostics

oktapam/fwprovider/resource_server_checkout_settings.go
```diff
@@ -0,0 +1,277 @@
...
+
+	// Call the SDK client to creat the server checkout settings from plan
+	serverCheckoutSettings := &pam.ResourceCheckoutSettings{
+		CheckoutRequired:          plan.CheckoutRequired,
+		CheckoutDurationInSeconds: plan.CheckoutDurationInSeconds,
+		IncludeList:               plan.IncludeList,
+		ExcludeList:               plan.ExcludeList,
+	}
+
+	_, err := r.client.SDKClient.ProjectsAPI.UpdateResourceGroupServerBasedProjectCheckoutSettings(ctx, r.client.Team, plan.ResourceGroup, plan.Project).ResourceCheckoutSettings(*serverCheckoutSettings).Execute()
```
sachinsaxena-okta, line 116: Create a separate file under oktapam/client to call SDK and handle errors there.

oktapam/fwprovider/resource_server_checkout_settings.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+	r.client = provider.SDKClientWrapper
+}
+
+func formatServerCheckoutSettingsID(resourceGroupID string, projectID string) string {
+	// server checkout settings don't have an identifier in itself and is really an attribute of a project.
+	// we manage it as a separate resource since it's lifecycle is somewhat separate from a project.
+	// since project password settings are managed as a separate resource with format resourceGroupID/projectID already,
+	// we can just append the server_checkout_settings to the end of the ID
+	return fmt.Sprintf("%s/%s/server_checkout_settings", resourceGroupID, projectID)
```
sachinsaxena-okta, line 276: You don't need to add server_checkout_settings. For a project and resource, there can be only one server checkout settings.

main.go
```diff
@@ -19,10 +24,54 @@ func main() {
...
+
+	// tf5to6server enables translating a protocol version 5 provider server into a protocol version 6 provider server.
+	upgradedSdkProvider, err := tf5to6server.UpgradeServer(
+		ctx,
+		oktapam.Provider().GRPCProvider,
+	)
+
+	// Combine Providers
+	// Refer: https://developer.hashicorp.com/terraform/plugin/mux/combining-protocol-version-6-providers
+	providers := []func() tfprotov6.ProviderServer{
```
sachinsaxena-okta, line 49: Move this code out to factory.go as it is used in both provider_test.go and main.go.

`func ProviderServerFactoryV6(ctx context.Context) (func() tfprotov6.ProviderServer, error) {

	v5Provider := oktapam.Provider()
	v6Provider := fwprovider.New()()

	// SDKV2 used for tf plugin development is designed for maintaining tf plugins that are compatible with Plugin
	// Protocol version 5. Plugins need to communicate with Terraform CLI, protocol version 5 is supported by CLI version
	// 0.12 and later. Protocol version 6 support tf cli version 1.0 or later.

	// To start using new TF Plugin Framework(https://developer.hashicorp.com/terraform/plugin/framework) we have two options -
	// Option 1: Downgrade new plugin framework server to support protocol version 5
	// Option 2: Upgrade old SDKV2 provider server to support protocol version 6
	// If we go with Option 1, then will not be able to use some of the newer features like Nested Attributes:
	// https://developer.hashicorp.com/terraform/plugin/framework/handling-data/attributes#nested-attribute-types

	//Going with option 2, that will require upgrading tf cli version to 1.0+.

	// tf5to6server enables translating a protocol version 5 provider server into a protocol version 6 provider server.
	upgradedV5Provider, err := tf5to6server.UpgradeServer(
		ctx,
		v5Provider.GRPCProvider,
	)

	if err != nil {
		return nil, err
	}

	// Combine Providers
	// Refer: https://developer.hashicorp.com/terraform/plugin/mux/combining-protocol-version-6-providers
	providers := []func() tfprotov6.ProviderServer{
		func() tfprotov6.ProviderServer {
			return upgradedV5Provider
		},
		providerserver.NewProtocol6(v6Provider),
	}

	muxServer, err := tf6muxserver.NewMuxServer(ctx, providers...)
	if err != nil {
		return nil, err
	}

	return muxServer.ProviderServer, nil
}`

sachinsaxena-okta: Now, in main.go you just need to do - 

`	muxServer, err := ProviderServerFactoryV6(ctx)
	if err != nil {
		log.Fatal(err)
	}

	var serveOpts []tf6server.ServeOpt

	if debug {
		serveOpts = append(serveOpts, tf6server.WithManagedDebug())
	}

	err = tf6server.Serve(
		"registry.terraform.io/okta.com/pam/oktapam",
		muxServer,
		serveOpts...,
	)`

oktapam/provider_test.go
```diff
@@ -143,3 +151,54 @@ func getTeamName() string {
...
+		fwProvider:    frameworkProvider,
+	}
+	return ctx, providers, muxServer
+}
+
+func testAccFrameworkMuxProvidersServer(ctx context.Context, sdkV2Provider *schema.Provider, fwProvider *fwprovider.OktapamFrameworkProvider) map[string]func() (tfprotov6.ProviderServer, error) {
+	return map[string]func() (tfprotov6.ProviderServer, error){
+		"oktapam": func() (tfprotov6.ProviderServer, error) {
+			// upgrade sdk v2 provider to protocol version 6
+			upgradedSdkProvider, err := tf5to6server.UpgradeServer(
```
sachinsaxena-okta, line 180: This code can we moved to oktapam/provider/factory.go as I mentioned in earlier comment.

oktapam/resource_server_checkout_settings_test.go
```diff
@@ -0,0 +1,186 @@
...
+	"testing"
+
+	"github.com/atko-pam/pam-sdk-go/client/pam"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/okta/terraform-provider-oktapam/oktapam/constants/attributes"
+)
+
+const testAccServerCheckoutSettingsCreateConfigFormat = `
+provider "oktapam" {}
```
sachinsaxena-okta, line 16: Create base config string with group, resource group and project. Then, reuse that for creating and updating settings.

oktapam/resource_server_checkout_settings_test.go
```diff
@@ -0,0 +1,186 @@
...
+	resource_group = oktapam_resource_group.test_acc_resource_group.id
+	ssh_certificate_type  = "CERT_TYPE_ED25519_01"
+	account_discovery     = true
+}
+`
+
+func createTestAccServerCheckoutSettingsDeleteConfig(dgaName, resourceGroupName, projectName string) string {
+	return fmt.Sprintf(testAccServerCheckoutSettingsDeleteConfigFormat, dgaName, resourceGroupName, projectName)
+}
+func TestAccServerCheckoutSettingsSource(t *testing.T) {
```
sachinsaxena-okta, line 93: Add some more test for below cases - 

1. First, removing exclude_list and changing it to include_list
2. Then, remove include_list and use just checkout_required field
3. Provide both include and exclude list
4. Update project/resource group

oktapam/fwprovider/resource_server_checkout_settings.go
sachinsaxena-okta: I think better to call read after creating the resource so that state will be set based on the returned data. We are following this pattern in other resources too.

oktapam/acctest.go
sachinsaxena-okta, line 1: Moved all the acceptance test utility methods to this file instead of keeping those in provider_test.go

oktapam/fwprovider/resource_server_checkout_settings.go
guoliangye-okta: Removed the redundant check.

I didn't find any mention in the official document suggesting to perform the Read operation inside the Create. I came across an interesting issue related to this topic. https://github.com/hashicorp/terraform-plugin-framework/issues/277. I believe we can eliminate the need for a Read operation after Create/Update with the new Terraform provider framework.

Dockerfile.ci
```diff
@@ -2,7 +2,7 @@ FROM golang:1.21-bullseye
 
 ARG APP_PATH
 
-RUN go install github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs@v0.7.0
+RUN go install github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs@v0.19.4
```
guoliangye-okta, line 5: Update the version to latest to get the docs generated correctly.

I've attempted several methods to resolve this error [https://github.com/okta/terraform-provider-oktapam/actions/runs/11475253141/job/31932723253?pr=125](vscode-file://vscode-app/Applications/Cursor.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html), but haven't had any success. Any thoughts?

README.md
```diff
@@ -69,6 +69,39 @@ If you are using terraform 0.12.x, you will need to run `make link_legacy`.  Thi
 $ make link_legacy
 ```
 
+### Developing with Terraform Provider Framework
```
guoliangye-okta, line 72: Add the steps to adding new resource/data resource with new framework

oktapam/sdkv2_to_plugin_framework_migration_test.go
guoliangye-okta, line 1: migration test with the new framework to ensure that it doesn't break the provider.

oktapam/resource_server_checkout_settings_test.go
guoliangye-okta, line 1: Acceptance test for new added resource with new framework

oktapam/resource_server_checkout_settings.go
```diff
@@ -0,0 +1,277 @@
...
+	resp.Schema = schema.Schema{
+		Description: descriptions.ResourceServerCheckoutSettings,
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Computed: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"checkout_duration_in_seconds": schema.Int32Attribute{
```
guoliangye-okta, line 59: platform requires this field.

oktapam/resource_server_checkout_settings.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+
+	// Set state to empty data
+	state.Id = types.StringValue("")
+	diags = resp.State.Set(ctx, state)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	// should we delete the state here?
```
guoliangye-okta, line 245: should we delete the state here?

Dockerfile.ci
guoliangye-okta: fixed in https://github.com/okta/terraform-provider-oktapam/pull/125/commits/50ae212fc747868c2e5f18c57d3ace79b39027dd

oktapam/provider_factory.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+func V6ProviderServerFactory(ctx context.Context) (func() tfprotov6.ProviderServer, error) {
+	v5Provider := Provider()
+	v6Provider := New()()
+	// SDKV2 used for tf plugin development is designed for maintaining tf plugins that are compatible with Plugin
+	// Protocol version 5. Plugins need to communicate with Terraform CLI, protocol version 5 is supported by CLI version
+	// 0.12 and later. Protocol version 6 support tf cli version 1.0 or later.
+
+	// To start using new TF Plugin Framework(https://developer.hashicorp.com/terraform/plugin/framework) we have two options -
```
SajanAlexander-okta, line 20: these comments lay out some options, but do not specify what we chose and why - i can figure it out from the code below, but i am not sure if other devs that are unfamiliar with TF would understand it.

oktapam/provider_sdkv2.go
```diff
@@ -0,0 +1,170 @@
+package oktapam
```
SajanAlexander-okta, line 1: this file seemed to have been renamed from `provider.go`, but the history isn't linking.  can we move it back to `provider.go` so we can preserve history?

oktapam/resource_server_checkout_settings.go
SajanAlexander-okta: are there docs for the new provider framework that indicate what we should do here?

oktapam/sdkv2_to_plugin_framework_migration_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/okta/terraform-provider-oktapam/oktapam/constants/attributes"
+)
+
+func TestAccResourceGroupUpgradeFromSdkv2(t *testing.T) {
```
SajanAlexander-okta, line 11: Could you explain more what this test is actually doing by adding additional comments on the test itself?

terraform-provider-oktapam.iml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
SajanAlexander-okta, line 1: Is this for your editor?  If so, could you remove it?

SajanAlexander-okta (changes requested):

oktapam/provider_factory.go
guoliangye-okta: updated comment to indicate we decided to use the new terraform provider framework and which option we adopted here. For details how to add new resouce/data resource, I added the session in https://github.com/okta/terraform-provider-oktapam/blob/2923d112dc55b414dce55662e8de22fc1df4dfcb/README.md#L72

oktapam/sdkv2_to_plugin_framework_migration_test.go
guoliangye-okta: I added a test description to provide more detail about what has been tested

oktapam/provider_sdkv2.go
guoliangye-okta: Yes, it has been renamed from `provider.go`, which uses SDK version 2 with Protocol version 5. We made this change to separate the two providers and to allow `provider_factory.go` to manage their combination. I'm not quite sure what you mean by the "history" of this.

oktapam/resource_server_checkout_settings.go
guoliangye-okta: I’ve noticed a pattern in the SDKv2 provider resources where the ID is set to an empty string. There's a recommendation to skip calling the `RemoveResource()` method from the response state . The framework manages this logic automatically if there are no error diagnostics. https://developer.hashicorp.com/terraform/plugin/framework/resources/delete#recommendations Therefore, I believe we’re in good shape here.

SajanAlexander-okta:
@jamesatwill-okta - it looks like this change will be held up by your fix for the security policy optional field update.

SajanAlexander-okta (dismissed):

SajanAlexander-okta (approved):

guoliangye-okta (approved):

SajanAlexander-okta (approved):

PR #128, opened 2024-03-04
OKTA-649265 - implement secrets

SajanAlexander-okta:
implementation of PAM secrets


oktapam/client/secret_folder.go
```diff
@@ -215,7 +233,11 @@ func ConvertSecretOrFolderListResponseToSecretFolder(resp pam.SecretOrFolderList
...
 
-func ConvertResolveSecretOrFolderResponse(resp *pam.ResolveSecretOrFolderResponse) SecretFolder {
+func ConvertResolveSecretOrFolderResponseToSecretFolder(resp *pam.ResolveSecretOrFolderResponse) (SecretFolder, error) {
+	if resp.Type != nil && *resp.Type != pam.SecretType_FOLDER {
+		return SecretFolder{}, fmt.Errorf("resolved entity was not a secret folder.  id: %s", resp.Id)
+	}
+
 	var parentSecretFolderID *string
 
 	if len(resp.Path) != 0 {
```
jamesbuchan-okta, line 243: if we get down here and len is 0, that means it is a root level folder?

oktapam/client/secrets.go
```diff
@@ -0,0 +1,318 @@
...
+					} else {
+						list = append(list, secret)
+					}
+				} else {
+					list = append(list, ConvertSecretOrFolderListResponseToSecret(parentFolderPath, s))
+				}
+			}
+		}
+
+		linkHeader := httpResp.Header.Get("Link")
```
jamesbuchan-okta, line 286: Should all this go in a helper / utils place?  I know terraform likes copy/pasting code, but i imagine this could be shared with other places?

oktapam/data_source_secrets.go
```diff
@@ -0,0 +1,82 @@
...
+		attrs := make(map[string]any, 2)
+
+		attrs[attributes.ID] = secret.Secret.Id
+		attrs[attributes.Name] = secret.Secret.Name
+
+		secretsList[idx] = attrs
+	}
+
+	d.Set(attributes.Secrets, secretsList)
+	d.SetId(fmt.Sprintf("%s|%s|%s", resourceGroupID, projectID, path))
```
jamesbuchan-okta, line 79: weird id?

oktapam/data_source_secrets_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/okta/terraform-provider-oktapam/oktapam/constants/attributes"
+)
+
+func TestAccDatasourceSecretList(t *testing.T) {
```
jamesbuchan-okta, line 11: should we test the pagination somewhere?

oktapam/data_source_secrets_test.go
```diff
@@ -0,0 +1,143 @@
...
+			{
+				Config: initConfig,
+			},
+			{
+				Config: fmt.Sprintf("%s\n%s", initConfig, data1Config),
+				Check:  resource.TestCheckResourceAttr(dataSource1Name, fmt.Sprintf("%s.#", attributes.Secrets), "1"),
+			},
+			{
+				Config: fmt.Sprintf("%s\n%s", initConfig, data2Config),
+				Check:  resource.TestCheckResourceAttr(dataSource2Name, fmt.Sprintf("%s.#", attributes.Secrets), "2"),
```
jamesbuchan-okta, line 44: can you add a comment here to make it clearer what we're testing? I think the first one is 1 because it's getting an individual secret and the second is 2 because it's getting all secrets in the parent folder?

oktapam/client/secret_folder.go
SajanAlexander-okta: that's correct

oktapam/client/secrets.go
SajanAlexander-okta: I think the last time this came up, we said that each of these implementations is slightly different in ways that make it a bit cumbersome to have a generic function.  

Additionally, my preference would be to have the SDK handle this instead of the provider or any other code that uses it.

oktapam/data_source_secrets_test.go
SajanAlexander-okta: Based on the response to the comment above, my preference would be to move that into the SDK and not have to worry about things here.

oktapam/data_source_secrets_test.go
SajanAlexander-okta: Done

jamesbuchan-okta (approved):

PR #135, opened 2024-07-19
Fix args type bug

UlfatSayyed-okta:


oktapam/resource_sudo_command_bundle_test.go
```diff
@@ -56,8 +56,7 @@ func createTestAccSudoCommandBundleCreateConfig(name string) string {
 		structured_commands {
 			command       = "/bin/run.sh"
 			command_type  = "executable"
-			args_type     = "custom"
-			args          = "ls"
+			args_type     = "any"
```
shrirangadgaonkar-okta, line 59: can you add tests for all combinations?

Ie: if raw & dir then no args
if executable then if custom args then we should have args
and if other args type then no args!

shrirangadgaonkar-okta (commented):
please add that test and then verify locally with the updated main.tf file!

oktapam/resource_sudo_command_bundle_test.go
UlfatSayyed-okta: done!

shrirangadgaonkar-okta (approved):

PR #141, opened 2024-10-31
Compose from SDK pieces

jamesatwill-okta:
- update to Go 1.23
- update to newest terraform-plugin-framework/etc
- make resource_server_checkout_settings from an embedded struct


oktapam/resource_server_checkout_settings.go
```diff
@@ -31,13 +30,54 @@ type serverCheckoutSettingsResource struct {
...
-	Id                        types.String `tfsdk:"id"`
-	ResourceGroup             string       `tfsdk:"resource_group"`
-	Project                   string       `tfsdk:"project"`
-	CheckoutRequired          bool         `tfsdk:"checkout_required"`
-	CheckoutDurationInSeconds types.Int32  `tfsdk:"checkout_duration_in_seconds"`
-	IncludeList               []string     `tfsdk:"include_list"`
-	ExcludeList               []string     `tfsdk:"exclude_list"`
+	Id            types.String `tfsdk:"id"`
+	ResourceGroup string       `tfsdk:"resource_group"`
+	Project       string       `tfsdk:"project"`
```
jamesatwill-okta, line 35: Project, ResourceGroup, Id are part of the URL, whereas ResourceCheckoutSettings is $ref'd in the spec. I _think_ would could in theory generate everything from the spec, but that feels like it's filled with sharp edges.

Instead I'm wondering if we autogenerated tfsdk versions of some entities, we could have a simpler* code generator to the ToSDK/FromSDK methods/functions?

sachinsaxena-okta: We can generate code from openapi spec using [terraform-plugin-codegen-openapi](https://github.com/hashicorp/terraform-plugin-codegen-openapi). It works but wherever we don't have a parity in structure between api and tf schema it will require some more customization. TF plugin generator also allow customizing mapping between OAS spec -> TF schema (https://github.com/hashicorp/terraform-plugin-codegen-openapi).

jamesatwill-okta: @sachinsaxena-okta I was playing with that originally, but their code generator is still pretty lacking. 1) It wants to take over the entire code generation process for the provider, and 2) it doesn't support polymorphism yet. 

It definitely seems like the best overall future solution, but yeah it requires parity.

oktapam/convert/resource_checkout_settings.go
```diff
@@ -0,0 +1,47 @@
+package convert
```
jamesatwill-okta, line 1: In theory, we could auto-generate this file.

oktapam/resource_server_checkout_settings.go
```diff
@@ -221,16 +218,14 @@ func (r *serverCheckoutSettingsResource) Delete(ctx context.Context, req resourc
 	}
 	defaultCheckoutDurationInSeconds := int32(900)
```
jamesatwill-okta, line 219: I'm almost certain this is the wrong way to do delete, but I'm not touching it.

oktapam/resource_server_checkout_settings_test.go
```diff
@@ -168,7 +168,7 @@ func TestAccServerCheckoutSettings(t *testing.T) {
 			// Update with both lists
 			{
 				Config:      createServerCheckoutSettingsUpdateWithBothListsConfig(delegatedAdminGroupName, resourceGroupName, projectName),
-				ExpectError: regexp.MustCompile(`400 Bad Request`),
+				ExpectError: regexp.MustCompile(`Only one of 'IncludeList' or 'ExcludeList' can be specified`),
```
jamesatwill-okta, line 171: New tf framework exposes the actual error message!

oktapam/resource_server_checkout_settings.go
sachinsaxena-okta: There is no delete endpoint for settings resource. So, it just update the setting to default values and disable it. It's kind of same as what's done for password_settings resource - https://github.com/okta/terraform-provider-oktapam/blob/24d3620f9cca314e1b18ca83ed6bdf92b03c2648/oktapam/resource_password_settings.go#L186-L185. Having some comment here will be better for clarity.

jamesatwill-okta: If we ever redo this, it'd be nice of "checkout settings" were an entity that just got attached to resources or projects or whatnot so it can be manipulated individually.

sachinsaxena-okta: You mean manage "checkout settings" as separate entity and then refer in project. It has include_list and exclude_list too which will be reference to resources applicable to the project. I feel in that case too will end up having separate settings for each resource.

sachinsaxena-okta (approved):

PR #142, opened 2024-11-02
Fix drift issue with security_policy resource

sachinsaxena-okta:
This PR fixes unexpected drift without any config changes for resource security_policy. Changes made - 

1. Earlier resource schema allow passing NamedObject model for sudo_command_bundles in principal_account_ssh privilege. NamedObject has 3 fields ID, Name and Type and all marked as optional fields. As part of api response all of the NamedObject fields are returned which create drift if only some of the fields are passed in the initial config. We can fix it by customizing the diff but in all other places in security policy tf schema, we just allow passing uuid only - [Resource Group Example](https://github.com/okta/terraform-provider-oktapam/blob/867d8f881d87891350b761493f1bea1e892848d2/oktapam/resource_security_policy.go#L46), [Principals Example](https://github.com/okta/terraform-provider-oktapam/blob/867d8f881d87891350b761493f1bea1e892848d2/oktapam/resource_security_policy.go#L58-L57.) 

For consistency, decided to only allow UUID field to refer sudo_command_bundles in security_policy resource.

2. Added test to make sure without any config changes, expect empty plan. We can enforce it for all the resources but for this PR just adding it for security_policy resource.


oktapam/client/security_policy.go
```diff
@@ -647,13 +647,9 @@ func (p *PrincipalAccountSSHPrivilege) ToResourceMap() map[string]any {
 		m[attributes.AdminLevelPermissions] = false
 	}
 	if len(p.SudoCommandBundles) > 0 {
-		scbs := make([]map[string]any, len(p.SudoCommandBundles))
```
sachinsaxena-okta, line 650: Updated api response mapping code. Map ID from the returned NamedObject for sudo_command_bundle.

oktapam/resource_security_policy_test.go
```diff
@@ -300,6 +301,16 @@ func TestAccSecurityPolicy(t *testing.T) {
 					),
 				),
 			},
+			{
+				// check for planned changes before applying any configurations
```
sachinsaxena-okta, line 305: Added test to make sure no plan update if config not changed.

oktapam/sdkv2_to_plugin_framework_migration_test.go
```diff
@@ -34,17 +35,12 @@ func TestAccResourceGroupUpgradeFromSdkv2(t *testing.T) {
...
-				// // If acceptance testing is still using terraform-plugin-sdk/v2,
-				// // use `PlanOnly: true` instead. When migrating to
-				// // terraform-plugin-testing, switch to `ConfigPlanChecks` or you
-				// // will likely experience test failures.
-				// ConfigPlanChecks: resource.ConfigPlanChecks{
-				// 	PreApply: []plancheck.PlanCheck{
-				// 		&plancheck.ExpectEmptyPlan{},
-				// 	},
-				// },
-				PlanOnly: true,
```
sachinsaxena-okta, line 47: PlanOnly is old way of doing it with sdk v2, so updated the test.

jamesatwill-okta (approved):
other than that one item, this makes sense to me, can take as-is, or re-approve after

SajanAlexander-okta (commented):
LGTM - can also approve if/when the nit from James is addressed

SajanAlexander-okta (approved):

PR #143, opened 2024-11-05
pin version for config

SajanAlexander-okta:
Pin the version for the config to be version two based on help article here: 

https://goreleaser.com/errors/version/

This was seen in the following run: https://github.com/okta/terraform-provider-oktapam/actions/runs/11686112558/job/32541126595#step:6:20

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,7 +42,7 @@ jobs:
         name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v3.0.0
         with:
-          version: latest
+          version: 2
           args: release --clean
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -42,7 +42,7 @@ jobs:
         name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v3.0.0
```
sachinsaxena-okta, line 43: Should we update this also to v6. Check example here - https://goreleaser.com/ci/actions/#workflow

sachinsaxena-okta: Based on the release notes with v6, by default it uses v2. 
https://goreleaser.com/blog/goreleaser-v2/#github-action

sachinsaxena-okta (approved):

PR #146, opened 2024-11-07
- small cleanup

jamesatwill-okta:


oktapam/resource_server_checkout_settings.go
```diff
@@ -96,31 +96,27 @@ func (r *serverCheckoutSettingsResource) Schema(_ context.Context, _ resource.Sc
 func (r *serverCheckoutSettingsResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
 	// Retrieve the server checkout settings values from the plan
 	var plan serverCheckoutSettingsResourceModel
-	diags := req.Plan.Get(ctx, &plan)
-	resp.Diagnostics.Append(diags...)
-	if resp.Diagnostics.HasError() {
+	if diags := req.Plan.Get(ctx, &plan); diags.HasError() {
```
jamesatwill-okta, line 99: follow the if err: =..;err != { .. } pattern

oktapam/resource_server_checkout_settings.go
```diff
@@ -28,7 +27,8 @@ func NewServerCheckoutSettingsResource() resource.Resource {
 }
 
 type serverCheckoutSettingsResource struct {
-	client *client.SDKClientWrapper
+	api      *pam.ProjectsAPIService
```
jamesatwill-okta, line 30: We actually don't need the sdkclientwrapper.. we just need the projectsapi

guoliangye-okta: We used the SDK client wrapper here to add consistent error handling and manage common HTTP status codes in a standardized way, similar to the old client. Is there any reason to revert to using the SDK API directly?

guoliangye-okta (approved):

oktapam/resource_server_checkout_settings.go
jamesatwill-okta: This resource makes no use of any of those things.

PR #147, opened 2024-11-12
remove globals for test configuration setup

jamesatwill-okta:
- don't initialize acceptance test provider in `func init`
- don't store acceptance test client as a global


oktapam/provider_test.go
```diff
@@ -19,39 +18,24 @@ func TestProvider(t *testing.T) {
 		t.Fatalf("err: %s", err)
 	}
 }
+func TestProvider_impl(t *testing.T) {
+	var _ *schema.Provider = Provider()
+}
 
-var testAccSDKV2Providers map[string]func() (*schema.Provider, error)
```
sachinsaxena-okta, line 23: Any advantage of this change? We can also write a TestMain and initialize global variable there. Although I don't see that much issue in calling function to get the factory all the time, because it just return a factory and then test going to initialize the actual instance of the provider.

sachinsaxena-okta (dismissed):

oktapam/provider_test.go
jamesatwill-okta: I think the only downside to doing this is we go through more service tokens, so we might expose a memory leak or rate limit.

sachinsaxena-okta (approved):

PR #149, opened 2024-11-13
Introduce security_policy_v2 using terraform framework

jamesatwill-okta:
This PR introduces a `oktapam_security_policy_v2` Terraform resource that uses the new Terraform Plugin Framework.

It lays out a resource schema that maps tightly to the actual SDK and API for SecurityPolicy in `pam-sdk-go` and on our platform. 

This resource is not made to be backwards compatible with `oktapam_security_policy` which will continue to work for the forseeable future, but will not be gaining new bells and whistles.

This PR also introduces `httpmock` to write Terraform tests that don't require an environment up and running for simple cases.

### NOTE

The new resource is written in such a way that we should be able to replace it with generated code once we're happy it's working.


oktapam/resource_security_policy2.go
```diff
@@ -0,0 +1,144 @@
...
+	teamName string
+	api      *pam.SecurityPolicyAPIService
+}
+
+func NewSecurityPolicyResource() resource.Resource {
+	return &SecurityPolicyResource{}
+}
+
+func (s *SecurityPolicyResource) Metadata(_ context.Context, request resource.MetadataRequest, response *resource.MetadataResponse) {
+	response.TypeName = request.ProviderTypeName + "_security_policy_v2"
```
jamesatwill-okta, line 26: This resource is not backwards compatible with security_policy, but at the same time we have no active plans to get rid of security_policy, it's just that we'll add new features to this resource instead.

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.22.0
 toolchain go1.23.2
```
jamesatwill-okta, line 6: I don't expect this to be 100% right away, I'd like to get it merged so people can build  the missing chunks while I fill out missing attributes and do cleanups.

oktapam/convert/resource_checkout_settings.go
```diff
@@ -3,6 +3,11 @@ package convert
 import (
```
jamesatwill-okta, line 3: I'm not particularly enamoured with the package name `/convert/`

Makefile
```diff
@@ -3,7 +3,7 @@ NAMESPACE=pam
 NAME=oktapam
 BINARY=terraform-provider-${NAME}
 # On version changes, update tag-checks.yml
```
jamesatwill-okta, line 5: @leharbhatt-okta  / @SajanAlexander-okta what does this comment mean?

SajanAlexander-okta: TLDR; you can take out that comment

So I think the idea here was that `tags-checks.yml` would define a GHA which would check out different versions of the provider and run the integration tests.  In theory, this would help us find regressions where a change in platform would cause a regression in an older version of the provider.  The implementation never really got off the ground (i.e. there were actually no old versions defined in `tags-checks.yml`) and the action errored out each time it was run, which resulted in removing the file.

jamesatwill-okta: Got it. Removed.

oktapam/convert/resource_checkout_settings.go
guoliangye-okta: Considering the purpose of converting files, renaming it to `modelconv` would make it more concise and descriptive?

jamesatwill-okta: Hmm... Looking at the AWS provider, they call the processes Expand and Flatten, and the process of doing that "flexing", so I'm kinda leaning towards that.

shrirangadgaonkar-okta (approved):

PR #152, opened 2024-11-25
- fix diags handling

jamesatwill-okta:
Looking at the AWS provider, diags are not meant to be swallowed, but instead always returned. This sets that pattern up.


shrirangadgaonkar-okta (approved):
LGTM

PR #153, opened 2024-11-26
Cleanup panics, errors, and TODOs.

jamesatwill-okta:


oktapam/convert/security_policy_privileges.go
```diff
@@ -78,7 +76,8 @@ func SecurityPolicyRulePrivilegeContainerFromSDKToModel(ctx context.Context, in
 		out.PasswordCheckoutSSHPrivilege = privilege
 
 	} else {
-		panic("missing stanza in SecurityPolicyRulePrivilegeContainerFromSDKToModel")
+		diags.AddError("missing stanza in OktaPAM provider", "missing stanza in SecurityPolicyRulePrivilegeContainerFromSDKToModel")
```
sachinsaxena-okta, line 79: I don't quite understand this error. This is when SDK -> Model so API response is missing the required field but error says missing stanza in oktapam provider.

oktapam/convert/security_policy_privileges.go
```diff
@@ -118,7 +117,8 @@ func SecurityPolicyRulePrivilegeContainerFromModelToSDK(ctx context.Context, in
 		privilegeType = pam.SecurityPolicyRulePrivilegeType_PASSWORD_CHECKOUT_SSH
 
 	} else {
-		panic("missing stanza in SecurityPolicyRulePrivilegeContainerFromModelToSDK")
+		diags.AddError("missing stanza in OktaPAM provider", "missing stanza in SecurityPolicyRulePrivilegeContainerFromModelToSDK")
```
sachinsaxena-okta, line 120: I think schema validation will mostly capture these types of error as it's marked as `Required` in schema. Right?

sachinsaxena-okta: I got it now by going over schema. So, privilege container is required but content inside that can be missing.

sachinsaxena-okta: As this error is user facing, I think we should refine this error. Something like - 

diags.AddError("invalid configuration for privileges[1]", "Privilege is empty. Each privilege must define exactly one type ('Type1', 'Type2'.)")

oktapam/convert/security_policy_privileges.go
jamesatwill-okta: If we add a new privilege value, and a human changes the privileges in the UI for an existing policy to one not supported by the Terraform provider, the provider won't know how to convert it to the model.

oktapam/convert/security_policy_privileges.go
jamesatwill-okta: It is user-facing, but it only happens if one of our developers doesn't add an extra stanza here to convert a privilege from terraform back to the sdk. Because it's a programming mistake, I was tempted to leave them as `panic`, but TF has a nice enough error handling system (with diagnostics) that I felt letting TF cleanup properly on an error was probably a better bet.

sachinsaxena-okta: This error will also come if let's say you have config like this below - 

```
      privileges = [
        {
        }, 
        {
          principal_account_ssh = {
            principal_account_ssh   = true
            admin_level_permissions = true
          }
        }
      ]
```

sachinsaxena-okta (approved):

